PR #2, opened 2020-06-24
Create homepage design, with basic sign in / sign out structure set up.

zpChris:
Create the homepage for TravIS, and show a sign in / sign out button that changes based on the user's current sign in status.

Bootstrap used to keep design consistent.


zpChris:
This is an image of what the site should look like: 
image

zpChris:
I removed the default push triggers after connecting this repository to Google Cloud Build, so all checks are now passing.

PR #3, opened 2020-06-24
Update design to populate homepage from JS.

zpChris:
Create functions to update webpage dynamically from JS. Prior to this CL, the design / structure was directly in the `index.html` file and the div containers had their `display` set to `none` in order to show the correct portions. Since this could allow a user to go into the inspect mode and show the signed-in view by changing the display of certain divs, I removed this content from the `index.html` file and conditionally inject this HTML components into `index.html` from `script.js` based on the user's sign in status.

I also created a function that builds the header to the site.

I am adding no reviewers until the earlier CL (branch: HomepageDesign) is approved.


zpChris:
The different pages should look as follows.

Homepage, not signed in: 
image

Homepage, signed in:
image

Trips page, signed in (purposely not designed yet, and I am aware of the console error on this page):
image

If you go to the Trips page without signing in (by typing it in the browser), it will show the following -- this is fixed on the next CL, as the user will be redirected to the sign in page if they are signed out:
image

zpChris:
I am aware that Bootstrap provides "navbar" styling, and mine is custom; I made a custom style for this. I am happy to change this to use more Bootstrap, but at least for the prototype I think we should stick with the current model.

src/main/webapp/script.js
```diff
@@ -38,12 +38,123 @@ function isSignedIn() {
...
+
+  // Add the header as the first child of the content container.
+  contentContainer.prepend(header);
 }
+
+// Set the width of the content container.
+function setContentWidth(width) {
+  const contentContainer = document.getElementById('content');
+  contentContainer.style.width = width;
+}
```
adamfweidman, line 160: nit: add line

adamfweidman: a couple other files also require a line at the end

zpChris: Done!

src/main/webapp/script.js
```diff
@@ -38,12 +38,123 @@ function isSignedIn() {
...
+function displayStartTripDesign() {
+  // Set the width of the content container.
+  setContentWidth('800px');
+
+  // Remove background image id from body.
+  document.body.removeAttribute('id');
+
+  // Get the homepage "start trip" block to and add elements.
+  const indexStartTripBlock = document.getElementById('index-start-trip-block');
+  indexStartTripBlock.style.display = 'block';
```
tristan-jh, line 98: If I start the page an click sign-in then the trips button I see the following error in the js console coming from this line:
script.js:26 TypeError: Cannot read property 'style' of null
    at displayStartTripDesign (script.js:98)
    at script.js:20

I think because clicking on the trips button redirects to /trips the 'index-start-trip-block' element no longer exists when this method is being called. But I'm not certain.

src/main/webapp/script.js
```diff
@@ -38,12 +38,123 @@ function isSignedIn() {
...
+  titleElement.innerText = 'TravIS: Travel Information System';
+  titleElement.className = 'index-registration-block-child';
+
+  // Create p element to describe TravIS.
+  const textElement = document.createElement('p');
+  textElement.innerText = 
+    'TravIS, your personal Travel Information System, allows you to plan ' +
+    'the optimal trip to wherever you want to visit.';
+  textElement.className = 'index-registration-block-child';
+  textElement.id = 'index-registration-block-p';
```
tristan-jh, line 64: Not required for the prototype this week. But I'd prefer to have these tags defined in the html page then shown/hidden by the javascript. Ideally there's division of responsibility between the html and javascript files. Where the html file should describe all the _contents_ of the page and the js file describes only the _behaviors_ of the page. Creating and adding all these tags here makes it harder to interpret the behavior (ie: what triggers a change what that change does) and harder to interpret what shows up on the page since this format is less readable (imo) than straight html.

I get that in this particular case it's a bit awkward because you don't know what to show while you're waiting for a response from the authserver. You can default to showing a blank or loading page and then show/hide the appropriate content based on how the authserver responds.

tristan-jh: Just saw your comment in the header.

I wouldn't worry too much about users seeing signed-in html content before they actually sign in, since that signed-in content won't reveal any users private information b/c you'd have to retrieve personal contents from the server first anyways. Plus this approach doesn't necessarily fix the problem b/c a user could view the script.js in the inspect tab as well as the html.

I'm not totally sure what the best strategy is for creating a foolproof sign-in barrier. This might be something to bring up with the VSE's. Maybe you could have a separate sign-in page and just re-direct to there if a user isn't signed in? or maybe there's other ways to conditionally load page content that doesn't require building it all up in js? Depending on the complexity of the problem this might be worth adding to the design doc too.

src/main/webapp/script.js
zpChris: Yes, I was aware of that and I fixed it in my next CL -- I'll try to remove this issue from this CL as well.

src/main/webapp/script.js
zpChris: Talked with Tristan on call -- decided to revert this back to HTML and use the JS just to change the `display` settings of different divs, depending on the user's sign-in status.

src/main/webapp/script.js
zpChris: Fixed!

src/main/webapp/script.js
zpChris: One option may be to have the content be fetched from a servlet for the index page; this would prevent the user from seeing any content they were not supposed to see. However, for the prototype, changing the `display` status in the JS file should be sufficient.

zpChris:
I've updated this CL based on @tristan-jh's comment, reverting to previous CL style by putting HTML content directly into `index.html`, and just using the `script.js` file for updating the `display` attribute of different HTML `div` elements. This is fine for the prototype as we are less concerned about malicious actors, though for the final product we may consider other options (such as fetching the HTML content from the server).

src/main/webapp/trips/index.html
```diff
@@ -0,0 +1,38 @@
...
+        </div>
+      </div>
+      <div id="index-registration-block">
+        <h1 class="index-registration-block-child">TravIS: Travel Information System</h1>
+        <p class="index-registration-block-child" id="index-registration-block-p">
+          TravIS, your personal Travel Information System, allows you to plan 	
+          the optimal trip to wherever you want to visit.	
+        </p>	
+        <form action="/auth" method="POST" class="index-registration-block-child"	
+          id="sign-in-button-form">
```
tristan-jh, line 30: I think this page will probably change to be different from index.html fairly soon. But if not, or if those pages will have some parts that are identical long-term, then you might consider looking for js libraries that would let you factor out the common parts (eg: handlebars.js i think? there may be others). Definitely not something to worry about on this CL though.

PR #5, opened 2020-06-26
Create mock start trip form on index page, and trips page that shows list of current trips.

zpChris:
I created a mockup of the "start trip" page (shown when user is signed in on index page) and the list view of trips at `/trips`; this is slides two, three, and four of the [TravIS Mockup](https://docs.google.com/presentation/d/1lxoAexLReI8X7Lngnj0fe5brIgL7S7bo3BDqf-kCvvY/edit#slide=id.g89200f5d5b_1_9). 
The "start trip" page is not dynamic yet' this is intentional, and I am planning on adding the following dynamic components on the next CL (you don't have to read this list, just for reference):
1. The text fields will show the "green checkmark" when a valid input is entered (`is-valid` class in Bootstrap).
2. The "End Date" field will be automatically filled with the same date as "Start Date", as our initial trips are just one day.
3. The "Point of Interest" and "Calculate Trip" groups will only appear once the first three fields (Destination and Dates) are filled, and the next button is pressed. The next button will initially be green, and then after press will turn a different color and will change to "Back", which can be clicked to re-edit the first three fields.
4. The "Add POIs" button will actually work.
5. The "click to remove" feature of the POIs will actually work.
6. The "Calculate Trip" button will be disabled until all fields are filled-in, and at least one POI is entered.
Later, for the MVP, the "Destination" and "Point of Interest" fields will have Google Location autofill. Eventually, we will try to suggest POIs based on the current "Destination".

I set up the `TripServlet.java` file, with the `/calculate-trip` URL (even though this is a prototype), because I wanted to ensure that this form design would actually enable me to get the user input upon submit. Eventually, the "Calculate Trip" button will actually calculate the trip and take the user to the Maps or Calendar page for that trip.

Upon reflection, I should've split up this CL into editing the `trips` page and the `index` page separately; please let me know if I can help clarify anything on this CL.

Note: I previously made the `script.js` file generate the HTML, but I have since reverted this after a discussion with Tristan; instead, I simply change the `display` attribute of different `div` elements from `script.js`. While this allows malicious users to view the logged-in view without being logged in, this is not a large concern for the prototype.


zpChris:
These are the screenshots that show what the site should look like now:

Homepage, before sign in (unchanged from last CL):
image

Homepage, after sign in (**changed from last CL**):
image

Trips page (`/trips/`), after sign in -- before sign in, redirects back to sign in page (**changed from last CL**):
image

Homepage result of clicking "Calculate Trip" button -- takes you to `POST` result of servlet at `/calculate-trip` (**created in this CL**):
image
(If you input a destination and start date, they will appear in this servlet view. If you visit this page without clicking "Calculate Trip," you will get an error: `Error 405 HTTP method GET is not supported by this URL` -- this is intentional, at least for the prototype.)

src/main/webapp/trips/tripScript.js
```diff
@@ -0,0 +1,49 @@
...
+}
+
+function displayHeader() {
+  // Display header for site.
+  const header = document.getElementById('header');
+  header.style.display = 'block';
+}
+
+// Set the width of the content container.
+function setContentWidth(width) {
```
eshika, line 46: There's some repeated code here with the functions in script.js (displayHeader() and setContentWidth(width)), so we should consider restructuring the code in a way that we don't have to define these same functions in every script file. But I think this is fine for the prototype!

zpChris: Absolutely agree, I'm looking to clean up some of the repeated JS code across files next week when building the MVP. Thanks for the note, this sounds good!

src/main/webapp/index.html
```diff
@@ -32,6 +32,43 @@ <h1 class="index-registration-block-child">TravIS: Travel Information System</h1
...
+          <div class="form-row">
+            <div class="form-group col-md-3">
+              <input type="text" placeholder="Point of Interest" 
+                class="form-control" id="inputPoi">
+            </div>
+            <div class="form-group">
+              <button type="button" class="btn btn-primary">Add POIs</button>
+            </div>
+            <div class="form-group">
+              <input type="button" value="California (click to remove)"
```
bheff-work, line 62: I would probably have an 'x' or edit ability, rather than have "(click to remove)" on the button. But since this is so early, you don't need to do anything about this right now. Just wanted to address it.

src/main/webapp/trips/tripScript.js
```diff
@@ -0,0 +1,49 @@
...
+// limitations under the License.
+
+// Triggered upon DOM load.
+$(document).ready(() => {
+  // Redirect to homepage if user is not signed in.
+  isSignedIn().then((signInStatus) => {
+    if (!signInStatus) {
+      window.location.replace('/');
+    } else {
+      setContentWidth('800px');
```
bheff-work, line 22: another comment that you don't really need to act on right now, but it's good to think about. Be wary of hard-coding pixel sizes, since they'll look differently on different sized windows.

But, again, just food for thought

zpChris: Agreed, I'll keep that in mind -- maybe I can base this on screen size (some percentage of the width) in the future.

src/main/webapp/index.html
zpChris: I agree with that as well -- I tried to add an "X" button for removal at first, but the design was a bit difficult so I stuck with this. Once we finish the prototype, I'll be sure to update this!

PR #9, opened 2020-06-29
Add JavaScript functions to make homepage "start trip" form dynamic.

zpChris:
This CL makes the "start trip" form dynamic. The list of changes are below:
1. The text fields will show the "green checkmark" when a valid input is entered (is-valid class in Bootstrap).
2. The "End Date" field will be automatically filled with the same date as "Start Date", as our initial trips are just one day.
3. The "Point of Interest" and "Calculate Trip" groups will only appear once the first three fields (Destination and Dates) are filled, and the next button is pressed. The next button will initially be green, and then after press will turn a different color and will change to "Back", which can be clicked to re-edit the first three fields.
4. The "Add POIs" button actually works as expected, by adding a POI. This button is disabled until a valid POI is entered. 
5. The "click to remove" feature of the POIs will actually work.
6. The "Calculate Trip" button will be disabled until all fields are filled-in, and at least one POI is entered.
7. Initially, only the location and date forms are shown (POI and submit button are hidden), until the "Next" button is pressed. When the "Next" button is pressed, the POI and submit button are shown, the location and date forms are changed to read-only, and the "Next" button turns into a "Back" button -- when pressed, the "Back" button switches back to the original view.

During the MVP, one of my next steps will be to make the "Destination" and "Point of Interest" fields have Google Location autofill. Eventually, we will try to suggest POIs based on the current "Destination".

### DIFF
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -35,22 +35,28 @@
         <form action="/calculate-trip" method="POST" id="startTripForm">
           <div class="form-row">
             <div class="form-group col-md-3">
-              <input type="text" placeholder="Destination" class="form-control" 
-                name="inputDestination" id="inputDestination">
+              <label for="inputDestination">Destination</label>
+              <input type="text" placeholder="Where to?" class="form-control" 
+                name="inputDestination" id="inputDestination" 
+                onchange="checkValidLocation('inputDestination'); checkNextButton()">
             </div>
             <div class="form-group col-md-3">
-              <input type="text" placeholder="Start Date" class="form-control" 
-                name="inputStartDate" id="inputStartDate">
+              <label for="inputStartDate">Start Date</label>
+              <input type="date" class="form-control" name="inputStartDate" 
+                id="inputStartDate" oninput="setEndDateValue()"
+                onchange="checkValidDate('inputStartDate'); checkNextButton()">
             </div>
             <div class="form-group col-md-3">
-              <input type="text" placeholder="End Date" class="form-control is-valid" 
-                name="inputEndDate" id="inputEndDate" readonly>
+              <label for="inputEndDate">End Date</label>
+              <input type="date" class="form-control" name="inputEndDate" 
+                id="inputEndDate" readonly>
             </div>
             <div class="form-group">
-              <button type="button" class="btn btn-secondary">Next</button>
+              <button type="button" class="btn btn-success" id="next-destination-dates"
+                 onclick="toggleStartTripInputStage()">Next</button>
             </div>
           </div>
-          <div class="form-row">
+          <div class="form-row" id="add-pois-container">
             <div class="form-group col-md-3">
               <input type="text" placeholder="Point of Interest" 
                 class="form-control" id="inputPoi">
@@ -67,7 +73,8 @@
                 class="btn btn-secondary poi-input" name="Oregon">
             </div>
           </div>
-          <input type="submit" class="btn btn-success" value="Calculate Trip">
+          <input type="submit" class="btn btn-success" id="submit-calculate-trip" 
+            value="Calculate Trip">
         </form>
       </div>
     </div>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -12,7 +12,7 @@
   <body>
     <div id="content">
       <div id="header">
-        <h3 class="header-child float-left" id="header-title">TravIS</h3>
+        <a href="/"><h3 class="header-child float-left" id="header-title">TravIS</h3></a>
         <form class="header-child float-right" id="sign-out-button-form" action="/auth" method="POST">
           <input class="btn btn-primary" type="submit" value="Sign out">
         </form>
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -38,13 +38,15 @@
               <label for="inputDestination">Destination</label>
               <input type="text" placeholder="Where to?" class="form-control" 
                 name="inputDestination" id="inputDestination" 
-                onchange="checkValidLocation('inputDestination'); checkNextButton()">
+                onchange="checkValidLocation('inputDestination'); checkNextButton();
+                  checkSubmitButton()">
             </div>
             <div class="form-group col-md-3">
               <label for="inputStartDate">Start Date</label>
               <input type="date" class="form-control" name="inputStartDate" 
                 id="inputStartDate" oninput="setEndDateValue()"
-                onchange="checkValidDate('inputStartDate'); checkNextButton()">
+                onchange="checkValidDate('inputStartDate'); checkNextButton(); 
+                  checkSubmitButton()">
             </div>
             <div class="form-group col-md-3">
               <label for="inputEndDate">End Date</label>
@@ -62,7 +64,8 @@
                 class="form-control" id="inputPoi">
             </div>
             <div class="form-group">
-              <button type="button" class="btn btn-primary">Add POIs</button>
+              <button type="button" class="btn btn-primary"
+                onclick="checkSubmitButton()">Add POIs</button>
             </div>
             <div class="form-group">
               <input type="button" value="California (click to remove)" 
--- a/src/main/webapp/index.html
+++ b/src/main/webapp/index.html
@@ -52,8 +52,8 @@
                 id="inputEndDate" readonly>
             </div>
             <div class="form-group">
-              <button type="button" class="btn btn-success" id="next-destination-dates"
-                 onclick="toggleStartTripInputStage()">Next</button>
+              <input type="button" class="btn btn-success" value="Next" 
+                id="toggle-stage-button" onclick="toggleStartTripInputStage()">
             </div>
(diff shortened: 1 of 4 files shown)


zpChris:
The following are screenshots of the homepage once the user is signed in, which shows the "start trip" form:

Basic homepage, no action taken:
image

Homepage after Destination and Start Date are entered (the "Next" button is now enabled):
image

Homepage after the "Next" button is pressed (the POI input and submit button are now shown, the "Next" button shows "Back", and the Destination and Date forms are now readonly):
image

Homepage after "Golden Gate Bridge" is added ("Calculate Trip" is now enabled):
image

POST request to `calculate-trip`; all inputs are shown (this is what will be replaced with an algorithm, eventually):
image

src/main/webapp/script.js
```diff
@@ -108,3 +125,182 @@ function addHiddenPoiFormTrigger() {
...
+
+// Build input attribute (for button); has type, value, className, name params.
+function buildInput(type, value, className, name) {
+  const form = document.createElement('input');
+  form.type = type;
+  form.value = value;
+  form.className = className;
+  form.name = name;
+  return form;
+}
```
adamfweidman, line 306: add line :)

src/main/webapp/script.js
```diff
@@ -108,3 +125,182 @@ function addHiddenPoiFormTrigger() {
...
+  } else {
+    // Hide the "Add POIs" form and "Submit" button.
+    const addPoiContainer = document.getElementById('add-pois-container');
+    addPoiContainer.style.display = 'none';
+    const startTripSubmitButton = document.getElementById('submit-calculate-trip');
+    startTripSubmitButton.style.display = 'none';
+
+    // Change location and date inputs to be editable.
+    const destinationInput = document.getElementById('inputDestination');
+    destinationInput.readOnly = false;
```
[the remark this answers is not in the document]
zpChris: `readOnly` means that the text input can't be edited anymore; thus, setting `readOnly` to `false` means you are letting them be editable again.

src/main/webapp/script.js
zpChris: Done!

src/main/webapp/script.js
```diff
@@ -108,3 +125,182 @@ function addHiddenPoiFormTrigger() {
...
+  if (isLocationDateInputValid()) {    
+    toggleStartTripStageButton.disabled = false;
+  } else {
+    toggleStartTripStageButton.disabled = true;
+  }
+}
+
+// If location and date forms have valid inputs, and there is at least one POI,
+// enable submit button; otherwise, disable button.
+function checkSubmitButton() {
```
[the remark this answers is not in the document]
zpChris: I'll absolutely do more "hover" displays in the future; right now, my priority is on getting the basic functionality out.

src/main/webapp/script.js
```diff
@@ -108,3 +125,182 @@ function addHiddenPoiFormTrigger() {
...
+function checkValidDate(elementId) {
+  const dateInput = document.getElementById(elementId);
+
+  // If an input exists, add 'is-valid' class.
+  if (dateInput.value !== '') {
+    dateInput.classList.add('is-valid');
+  } else if (dateInput.classList.contains('is-valid')) {
+    dateInput.classList.remove('is-valid');
+  }
+}
```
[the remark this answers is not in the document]
zpChris: Done -- thanks!

src/main/webapp/index.html
```diff
@@ -35,39 +35,42 @@ <h1 class="index-registration-block-child">TravIS: Travel Information System</h1
...
+                onchange="checkValidInput('inputStartDate'); checkNextButton(); 
+                  checkSubmitButton()">
             </div>
             <div class="form-group col-md-3">
-              <input type="text" placeholder="End Date" class="form-control is-valid" 
-                name="inputEndDate" id="inputEndDate" readonly>
+              <label for="inputEndDate">End Date</label>
+              <input type="date" class="form-control" name="inputEndDate" 
+                id="inputEndDate" readonly>
             </div>
```
tristan-jh, line 55: I'd argue you should hide the end-date input entirely from the UI until multi-day trips are ready. And maybe rename start-date to 'day of travel' or similar. You and I know this projects MVP deals just with a single days itinerary but a user who doesn't know that might be confused why they can't change the end date.

src/main/webapp/index.html
```diff
@@ -35,39 +35,42 @@ <h1 class="index-registration-block-child">TravIS: Travel Information System</h1
...
             <div class="form-group">
-              <button type="button" class="btn btn-secondary">Next</button>
+              <input type="button" class="btn btn-success" value="Next" 
+                id="toggle-stage-button" onclick="toggleStartTripInputStage()">
             </div>
           </div>
-          <div class="form-row">
+          <div class="form-row" id="add-pois-container">
```
tristan-jh, line 61: Optional: my personal preference would be for the POI's I've already entered to stack up vertically instead of line-wrapping as they currently do. Eg vertically stacking these elements: https://screenshot.googleplex.com/whUAXWbqF58

src/main/webapp/index.html
```diff
@@ -35,39 +35,42 @@ <h1 class="index-registration-block-child">TravIS: Travel Information System</h1
...
-                class="form-control" id="inputPoi">
-            </div>
-            <div class="form-group">
-              <button type="button" class="btn btn-primary">Add POIs</button>
-            </div>
-            <div class="form-group">
-              <input type="button" value="California (click to remove)" 
-                class="btn btn-secondary poi-input" name="California">
+              <input type="text" placeholder="Point of Interest" id="inputPoi" 
+                class="form-control" onchange="checkValidInput('inputPoi');
```
tristan-jh, line 64: Optional: So for me the workflow was to type the POI i'm interested in then click elsewhere on the page, then click on the add button. I'd prefer to just have the add button become clickable once I've typed enough characters, so that I can skip that first click. If I'm understanding correctly this is because the onchange event is only issued when the input box changes focus (ie, when I click away).

This is a small issue, and I suspect some of this UI will change once you start integrating the maps POI suggestion API. So probably easiest to deal with this later.

src/main/webapp/script.js
```diff
@@ -108,3 +125,169 @@ function addHiddenPoiFormTrigger() {
...
+ * If toggle "Next/Back" button equals "Next", set location and date inputs to
+ * readonly and show POI input and submit button.
+ * If toggle "Next/Back" button equals "Back", set location and date inputs to 
+ * editable and hide POI input and submit button.
+ */
+function toggleStartTripInputStage() {
+  const toggleStartTripStageButton = document.getElementById('toggle-stage-button');
+
+  if (toggleStartTripStageButton.value === 'Next') {
+    // Display the "Add POIs" form and "Submit" button.
```
tristan-jh, line 219: No action required: Mostly a note-to-self but we should discuss the model-view-controller pattern. Bootsie may know it better than I do, but it's a strategy for separating the state of the webpage from the visual display of the webpage. Mentioning it here because it would involve changing this check, so that instead of looking at the value of an HTML object you'd keep a "model" object in javascript which would keep track of what stage of input the user is in.

MVC might be overkill for this page, so no action requested in this PR, just something we should discuss and consider.

src/main/webapp/index.html
```diff
@@ -35,39 +35,42 @@ <h1 class="index-registration-block-child">TravIS: Travel Information System</h1
         <form action="/calculate-trip" method="POST" id="startTripForm">
           <div class="form-row">
             <div class="form-group col-md-3">
-              <input type="text" placeholder="Destination" class="form-control" 
-                name="inputDestination" id="inputDestination">
+              <label for="inputDestination">Destination</label>
+              <input type="text" placeholder="Where to?" class="form-control" 
+                name="inputDestination" id="inputDestination" 
+                onchange="checkValidInput('inputDestination'); checkNextButton();
```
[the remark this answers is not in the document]
zpChris: Ok, I can do that -- does the same go for `onclick` functions of buttons?

src/main/webapp/index.html
zpChris: Done!

src/main/webapp/index.html
zpChris: Yes, I was aware of this -- you're right, this aspect of the UI changes once the Google Place autofill is added, so I'm going to leave this as-is in this CL. Thanks for the note though!

src/main/webapp/index.html
zpChris: I like that too -- done!

src/main/webapp/index.html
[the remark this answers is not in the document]
tristan-jh: yea same goes for onclick. I think the convention should be that all callbacks are established in the js file as opposed to directly in the html.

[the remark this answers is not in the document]
zpChris: I didn't see this until now, and merged the branch without doing this -- I will include this change in my next CL. Sorry about that, I'll make it clear when it's done on the future CL!

PR #15, opened 2020-07-01
Autofill place location for relevant text input fields.

zpChris:
Main changes in this CL:
1. Add Google Place location autofill; in order for the user to enter valid input, they must click on one of the autofill options.
2. Add a "Name of Trip" input the the start of the form.
3. Refine the logic in `script.js` for enabling and disabling buttons (I would attach a video, but I resolved some errors that were occurring when you tried to edit multiple times, or pressed "Next" then "Back").
4. Move `onclick` listeners from `index.html` to `script.js`.

Image of starting form layout:
image

Images of location autofill and "Name of Trip" field:

image
image


src/main/webapp/script.js
```diff
@@ -148,17 +163,24 @@ function addHiddenPoiFormTrigger() {
     const poiInputs = document.getElementsByClassName('poi-input');
     let count = 1;
     Array.prototype.forEach.call(poiInputs, (poiInput) => {
-      $('<input>').attr('type', 'hidden')
-      .attr('name', 'poi-' + count)
-      .attr('value', poiInput.name)
-      .appendTo('#startTripForm');
+      let poiNumber = 'poi-' + count;
+      addHiddenInput(poiNumber, poiNumber, poiInput.name);
```
bheff-work, line 167: not sure I understand why this should be a helper method, are you planning to use it elsewhere? Seems very specific to this situation, and I don't know if it increases the clarity. I was actually confused what you meant by 'hidden input' until I read the code itself.

Not a huge deal, it could be a personal preference. So if you feel strongly about keeping it this way, that's fine with me

src/main/webapp/script.js
```diff
@@ -245,11 +269,13 @@ function toggleStartTripInputStage() {
...
-    destinationInput.readOnly = true;
-    const dayOfTravelInput = document.getElementById('inputDayOfTravel');
-    dayOfTravelInput.readOnly = true;
+    // Change name, location, and date inputs to be readonly.
+    const inputTripName = document.getElementById('inputTripName');
+    inputTripName.readOnly = true;
+    const inputDestination = document.getElementById('inputDestination');
+    inputDestination.readOnly = true;
+    const inputDayOfTravel = document.getElementById('inputDayOfTravel');
+    inputDayOfTravel.readOnly = true;
```
bheff-work, line 278: I would argue that putting all these elements into their own variables is actually starting to decrease your clarity here. Take 'inputTripName':

1. the name of the variable is the same as you're passing into getElementById, so you aren't actually adding to a user's comprehension of this element by adding a name
2. you aren't ever using this variable again, so it doesn't keep the line count down
3. "document.getElementById('inputTripName').readOnly = 'true'" isn't actually difficult to read and understand.

src/main/webapp/script.js
```diff
@@ -317,3 +345,39 @@ function buildInput(type, value, className, name) {
   form.name = name;
   return form;
 }
+
+// Get all location input fields, and add the Google Places autofill.
+function addLocationAutofill() {
+  let locationInputs = document.getElementsByClassName('places-autofill');
```
bheff-work, line 351: I'm only seeing 'places-autofill' used here, and since you're doing completely different things based on if it's 'inputDestination' or 'inputPOI', why not give them different class names?

src/main/webapp/script.js
zpChris: That makes sense -- done!

src/main/webapp/script.js
zpChris: This is a good point, I started to realize this as I kept writing these two lines in pairs. Fixed, thanks for the note!

src/main/webapp/script.js
zpChris: Ok, I removed the 'places-autofill' class and just used the IDs of the HTML elements instead. I initially used the 'places-autofill' class as there is lots of overlap (both enable location autocomplete and have the same 'is-valid' check), but this is OK because I add more input field-specific code to future CLs.

Done!

src/main/webapp/script.js
```diff
@@ -312,38 +313,44 @@ function buildInput(type, value, className, name) {
...
-      // For "input destination" field, check "Next" button; for POI, check 
-      // "Add POI" button.
-      if (locationInput.id === 'inputDestination') {
-        checkNextButton();
-      } else if (locationInput.id === 'inputPoi') {
-        checkAddPoiButton();
-      }
-    });
+// Add text input POI Google Places autofill.
+function addInputPoiLocationAutofill() {
```
bheff-work, line 317: You could parameterize this function probably to make it work for both inputPoi and inputDestination, but I'm not overly fussed. If you do decide to parameterize though, ask for me to re-review

PR #16, opened 2020-07-01
Create event class and eventTest class

adamfweidman:
I have created a basic event class where the purpose it to be able to easily parse the information into the fullCalendar fields:
1. It takes in a title (which will be the address), later in the pop up the name will be displayed and the address below it
2. it takes in a start and end date which have the format:  '2020-06-29T09:00:00' (I created functions that converted information into that format"
3. Lastly I create 2 constructors, one where no timeSpent is specified and is therefore set at default one hour and one where one can specify the time. For the moment we are assuming a default of one hour but later on depending on the time of event we will want to change the amount of time spent (EG: museum vs restaurant vs park etc)


src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+package com.google.sps.data;
+ 
+/**
+ * Class that creates an event specific to a POI 
+ */
+public class Event {
+ 
+  // inputs
+  private String name;
+  private String address;
```
zpChris, line 24: No action required: Eventually, we'll have a place object here, so you can get other information about the event to display on the calendar for post-MVP work. This is just a note that I wanted to put for confirmation.

adamfweidman: Yes absolutely!

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+
+  // fields to be saved for parsing
+  private String strStartTime;
+  private String strEndTime;
+ 
+  // class constants
+  private final static int HALFHOUR = 30;
+  private final static int HOUR = 60;
+  private final static int MINUTES_IN_A_DAY = 1440;
+  private final static int MIN_POSSIBLE_TIME = 0;
```
eshika, line 38: Nit: According to the style guide, the order of modifiers should be private static final

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+
+  /**
+   * Constructor that takes in time spent at location
+   * 
+   * @param name date from user input travel day. Cannot be null.
+   * @param address time to to start or end event. Must be within bounds
+   * @param date day of travel
+   * @param startTime start of activing
```
eshika, line 46: Nit: typo here? Did you mean activity?

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+   * @param travelTime time spent traveling to next location
+   * @param timeAtLocation time spent at POI
```
eshika, line 48: What are the units for these parameters? Hours or minutes?

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+ * Class that creates an event specific to a POI 
+ */
+public class Event {
+ 
+  // inputs
+  private String name;
+  private String address;
+  private String date;
+  private int startTime;
+  private int endTime;
```
zpChris, line 27: Can you add a comment somewhere explaining that these are in military time?

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+   * function that calculates endTime given start and timeSpent
+   * 
+   * @param timeAtLocation time that the user spends at a POI
+   */
+  private int calculateEndTime(int timeAtLocation) {
+    return this.startTime + Integer.parseInt(convertToFormat(timeAtLocation));
+  }
+
+  /**
+   * function that converts mintues into hhmm format
```
zpChris, line 123: nit: *minutes

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+  private final static int HALFHOUR = 30;
+  private final static int HOUR = 60;
+  private final static int MINUTES_IN_A_DAY = 1440;
+  private final static int MIN_POSSIBLE_TIME = 0;
+
+  /**
+   * Constructor that takes in time spent at location
+   * 
+   * @param name date from user input travel day. Cannot be null.
+   * @param address time to to start or end event. Must be within bounds
```
eshika, line 44: From what I understand the address field would store the address of the location, not the time? Let me know if I'm misinterpreting this field.

src/test/java/com/google/sps/EventTest.java
```diff
@@ -0,0 +1,117 @@
...
+    Assert.assertEquals(GOLDENGATEPARK, e1.getName());
+    Assert.assertEquals(ADDRESS, e1.getAddress());
+    Assert.assertEquals(DATE, e1.getDate());
+    Assert.assertEquals(START_TIME, e1.getStartTime());
+    Assert.assertEquals(DEFAULT_TRAVEL_TIME, e1.getTravelTime());
+    Assert.assertEquals(1100, e1.getEndTime());
+
+    Assert.assertEquals(STR_START_TIME, e1.getStrStartTime());
+    Assert.assertEquals(STR_DEF_END_TIME, e1.getStrEndTime());
+  }
```
zpChris, line 116: nit: I'd prefer adding a test to test that the constructor throws an error when you try to initialize an illegal event (i.e.; invalid date or time).

src/main/java/com/google/sps/data/Event.java
adamfweidman: Yes sorry these are typos. name: was going to be like Empire State Building as opposed to the address.
Address is full address

adamfweidman: Sorry this is my bad. Name is just the name: EG "empire state building" the address is the full address: Number street, city, etc...

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+    if (Integer.parseInt(numHours) < 10) {
+      numHours = "0" + numHours;
+    } 
+
+    if (Integer.parseInt(numMins) < 10) {
+      numMins = "0" + numMins;
+    }
+
+    return numHours + numMins;
+  }
```
zpChris, line 149: Why is this function `static`? In what situations will we use this outside the class? (Not saying it shouldn't be `static`, just genuinely curious.)

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,184 @@
...
+    return output;
+  }
+
+  /**
+   * function that calculates endTime given start and timeSpent
+   * 
+   * @param timeAtLocation time that the user spends at a POI
+   */
+  private int calculateEndTime(int timeAtLocation) {
+    return this.startTime + Integer.parseInt(convertToFormat(timeAtLocation));
```
eshika, line 119: Would this create issues if the minutes exceeds 60? For example if the startTime is 3:45 and timeAtLocation is 30 mins wouldn't this return 3:75? Just want to clarify what values you're adding here

adamfweidman: you are right

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,189 @@
...
+   * @param date date from user input travel day. Cannot be null.
+   * @param time time to to start or end event. Must be within bounds
+   */
+  private static String createStrTime(String date, int time) {
+    if (date == null) {
+      throw new NullPointerException("Date cannot be null");
+    }
+
+    if (time < MIN_POSSIBLE_TIME) {
+      throw new IllegalArgumentException("Time cannot be less than 0");
```
tristan-jh, line 103: nit: use 'MIN_POSSIBLE_TIME' constant in the message here so that if MIN_POSSIBLE_TIME is ever changed this exception message still makes sense.

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    if (time >= MINUTES_IN_A_DAY) {
+      throw new IllegalArgumentException("Time cannot be more than a 24hrs");
```
tristan-jh, line 107: same as above

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    String output = "";
+    output += date + "T" + Integer.toString(time).substring(0, 2) + ":" +
```
tristan-jh, line 111: Consider using a date parsing library like SimpleDateFormat [link: https://docs.oracle.com/javase/8/docs/api/java/text/SimpleDateFormat.html#parse-java.lang.String-java.text.ParsePosition-] to outsource this parsing logic.

Using a date library will also take care of edge cases like daylight savings time, leap years etc.

If your work is blocking on this CL I'd be ok with submitting it as-is, but I think I'd prefer to switch to a date library in a follow-up CL.

src/main/java/com/google/sps/data/Event.java
tristan-jh: A `static` function cannot access non-static fields of the enclosing class. But, otoh, a client can call static functions without an instance of the class. Eg: your code could call `Event.convertToFormat(1234)` without having to construct an Event object. So, as in this case, `static` is often used for stateless utility functions ("stateless" is another way of saying it doesn't modify class fields).

@adamfweidman can probably give more context on how/why he decided to have this be its own method and why he wanted it to be public.

src/test/java/com/google/sps/EventTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+  // time class constants
+  private final static int HOUR = 60;
+  private final static int HALF_HOUR = 30;
+  private final static int NINETY_MIN = 90;
+  private final static int HOUR_FIFTY_FIVE = 115;
+  private final static int THREE_HOURS = 180;
+
+  // files for event 1
+  private static final String GOLDENGATEPARK = "GGPark";
```
tristan-jh, line 36: nit: GOLDEN_GATE_PARK

src/test/java/com/google/sps/EventTest.java
```diff
@@ -0,0 +1,129 @@
...
+
+    Assert.assertEquals(GOLDENGATEPARK, e1.getName());
+    Assert.assertEquals(ADDRESS, e1.getAddress());
+    Assert.assertEquals(DATE, e1.getDate());
+    Assert.assertEquals(START_TIME, e1.getStartTime());
+    Assert.assertEquals(DEFAULT_TRAVEL_TIME, e1.getTravelTime());
+    Assert.assertEquals(1120, e1.getEndTime());
+
+    Assert.assertEquals(STR_START_TIME, e1.getStrStartTime());
+    Assert.assertEquals(STR_MANUAL_END_TIME, e1.getStrEndTime());
```
tristan-jh, line 99: nit: this may actually be easier to read if DATE, START_TIME, STR_START_TIME, and STR_MANUAL_END_TIME were not constants.

Styling guidance is pretty different for tests. Eg: in non-test code you'd want to eliminate the redundancy of declaring the same string in multiple places as you have done here. But in a test redundancy is ok if it improves clarity. So in this case if the strings were declared inline I'd have all the information I need to understand this test in one place, eg: I could see that the date part of STR_START_TIME is the same as DATE, etc without scrolling back and forth between this method and the constants declared above. (compare with the ADDRESS constant which is more appropriate since I can see it's given in the constructor and returned unmodified by the getter, so I don't really need to know it's contents in order to see what's going on)

Feel free to disregard this comment for now, I'll see if I can find some documentation on test styling that provides a bit more context

src/test/java/com/google/sps/EventTest.java
```diff
@@ -0,0 +1,129 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import com.google.sps.data.Event;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
```
tristan-jh, line 20: Great test coverage! I especially like the tests for invalid inputs.

src/test/java/com/google/sps/EventTest.java
adamfweidman: Ok! I thought so too but I thought styling said that I needed to declare all constants. LMK if you find anything and I'll change it!

src/main/java/com/google/sps/data/Event.java
adamfweidman: Would love to talk about this in over the phone tomorrow when you have the chance!

src/test/java/com/google/sps/EventTest.java
tristan-jh: Yea, styling for test code can be pretty different. Here's some documentation that another host put together: http://go/step-tott especially check out the 'Tests Too DRY? Make them DAMP' document which focuses on the redundancy vs legibility trade-off for tests.

adamfweidman:
Most recent commits add SimpleDateFormat to Event.java and I moved around the variables in EventTest.java to make the test more readable.

src/main/java/com/google/sps/data/Event.java
```diff
@@ -0,0 +1,199 @@
...
+    if (time >= MINUTES_IN_A_DAY) {
+      throw new IllegalArgumentException("Time cannot be more than " + MINUTES_IN_A_DAY);
+    }
+
+    String output = "";
+    output += date + "T" + Integer.toString(time).substring(0, 2) + ":" + 
+      Integer.toString(time).substring(2, 4) + ":00";
+
+    try {
+      SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
```
tristan-jh, line 117: optional: I think you can construct a SimpleDateFormat which more closely matches the input format and save the trouble of reformatting the string above, eg: `new SimpleDateFormat("yyyy-MM-ddHHmm")`. then pass in `date + Integer.toString(time)`.

adamfweidman: For the moment if all is working I think I'll keep what I have but I'll keep this in mind going forward!

PR #17, opened 2020-07-01
Create Trip and TripDay classes.

eshika:
In this CL, I have created the basic Trip and TripDay classes, which will be used to store information about a user's trips.

The Trip class has the following fields: tripName, startDate, endDate, numDays, and tripDays (an ArrayList of TripDay objects). I've created two constructors for Trip, one where the trip is limited to one day and the endDate is the same as the startDate (for the MVP), and one where all the fields can be set (for beyond the MVP).

The TripDay class has the following fields: origin, destination, and locations (an ArrayList of place IDs, which are strings). This is the basic information needed for generating a DirectionsRequest with the Google Maps API. 

Some notes on how the structure of these will change in the future:

1. The TripDay class will also have a field that contains an ArrayList of events (for the calendar), which I can add after Adam's CL #16 is approved and merged.
2. The Trip class will also have a user field that stores the user whose trip it is. Will work with Chris to add this.
3. The TripDay will also have fields for storing the resulting route and information calculated by the Maps API, but I'm still exploring the Maps API's Java client (for calculating routes in the Java servlets instead of the Javascript) to figure out how these can be stored.

### DIFF
--- a/src/main/java/com/google/sps/Trip.java
+++ b/src/main/java/com/google/sps/Trip.java
@@ -152,7 +152,9 @@ public class Trip {
    * Returns an List<TripDay> of TripDays for this trip.
    */
   public List<TripDay> getTripDays() {
-    return this.tripDays;
+    List<TripDay> tripDaysCopy = new ArrayList<>();
+    tripDaysCopy.addAll(this.tripDays);
+    return tripDaysCopy;
   }
 
   /**
--- a/src/main/java/com/google/sps/Trip.java
+++ b/src/main/java/com/google/sps/Trip.java
@@ -183,8 +183,8 @@ public class Trip {
       // is 1 more than the difference between the dates.
       int numDays = Math.round(difference / (1000*60*60*24)) + 1;
       return numDays;
-	 } catch (Exception e) {
+    } catch (Exception e) {
       throw new IllegalArgumentException("Invalid startDate or endDate.");
-	 }
+    }
   }
 }
--- a/src/main/java/com/google/sps/Trip.java
+++ b/src/main/java/com/google/sps/Trip.java
@@ -16,11 +16,16 @@ package com.google.sps;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Date;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
 
 /**
  * Trip is the class for storing a single trip (could be multiple days).
  */
 public class Trip {
+  private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
+
   private String tripName;
   private String startDate;
   private String endDate;
@@ -36,7 +41,7 @@ public class Trip {
    * @param numDays The number of days for the trip. 
    * @param tripDays The list of tripDays. Must be non-null.
    */
-  public Trip(String tripName, String startDate, String endDate, int numDays, List<TripDay> tripDays) {
+  public Trip(String tripName, String startDate, String endDate, List<TripDay> tripDays) {
     if (tripName == null) {
       throw new IllegalArgumentException("tripName cannot be null");
     }
@@ -57,10 +62,18 @@ public class Trip {
       throw new IllegalArgumentException("numDays must be an integer between 1 and 31.");
     }
 
+    if (!isValidDate(startDate)) {
+      throw new IllegalArgumentException("Invalid startDate format.");
+    }
+
+    if (!isValidDate(endDate)) {
+      throw new IllegalArgumentException("Invalid endDate format.");
+    }
+
     this.tripName = tripName;
     this.startDate = startDate;
     this.endDate = endDate;
-    this.numDays = numDays;
+    this.numDays = calcNumDays(this.startDate, this.endDate);
     this.tripDays = new ArrayList<>();
     this.tripDays.addAll(tripDays);
   }
@@ -130,5 +143,28 @@ public class Trip {
    */
   public List<TripDay> getTripDays() {
     return this.tripDays;
+  }
+
+  private static boolean isValidDate(String inDate) {
+    DATE_FORMAT.setLenient(false);
+    try {
+        DATE_FORMAT.parse(inDate.trim());
+    } catch (ParseException pe) {
+        return false;
+    }
+    return true;
+  }
+
+  private static int calcNumDays(String startDateString, String endDateString) {
+    try {
+	       Date startDate = myFormat.parse(startDateString);
+	       Date endDate = myFormat.parse(endDateString);
+	       long difference = endDate.getTime() - startDate.getTime();
+         // Convert milliseconds to days
+	       float numDays = (difference / (1000*60*60*24));
+	       return numDays;
+	 } catch (Exception e) {
+	       throw new IllegalArgumentException("Invalid startDate or endDate.");
+	 }
   }
 }
--- a/src/main/java/com/google/sps/Trip.java
+++ b/src/main/java/com/google/sps/Trip.java
@@ -115,7 +115,7 @@ public class Trip {
   }
 
   /**
-   * Returns an List<TripDay> copy of TripDays for this trip.
+   * Returns a List<TripDay> copy of TripDays for this trip.
    */
   public List<TripDay> getTripDays() {
     List<TripDay> tripDaysCopy = new ArrayList<>();
--- a/src/main/java/com/google/sps/Trip.java
+++ b/src/main/java/com/google/sps/Trip.java
@@ -15,15 +15,9 @@
 package com.google.sps;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.Set;
 
 /**
- * Event is the container class for when a specific group of people are meeting and are therefore
- * busy. Events are considered read-only.
+ * Trip is the class for storing a single trip (could be multiple days).
  */
 public class Trip {
   private String tripName;
@@ -33,15 +27,15 @@ public class Trip {
   private ArrayList<TripDay> tripDays;
 
   /**
-   * Creates a new event.
+   * Creates a new Trip.
    *
    * @param tripName The human-readable name for the trip. Must be non-null.
    * @param startDate The start date for the trip. Must be non-null.
    * @param endDate The end date for the trip. Must be non-null.
-   * @param numDays The number of days of the trip. Must be non-null.
+   * @param numDays The number of days for the trip. Must be non-null.
    * @param tripDays The list of tripDays. Must be non-null.
    */
-  public Trip(String tripName, String startDate, String endDate, ArrayList<TripDay> tripDays) {
+  public Trip(String tripName, String startDate, String endDate, int numDays, ArrayList<TripDay> tripDays) {
     if (tripName == null) {
       throw new IllegalArgumentException("tripName cannot be null");
     }
@@ -54,6 +48,10 @@ public class Trip {
       throw new IllegalArgumentException("endDate cannot be null");
     }
 
+    if (numDays == null) {
+      throw new IllegalArgumentException("numDays cannot be null");
+    }
+
     if (tripDays == null) {
       throw new IllegalArgumentException("tripDays cannot be null. Use empty array instead.");
     }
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    if (tripDays == null) {
+      throw new IllegalArgumentException("tripDays cannot be null. Use empty array instead.");
+    }
+
+    this.tripName = tripName;
+    this.startDate = startDate;
+    this.endDate = endDate;
+    this.numDays = numDays;
```
adamfweidman, line 58: I suggest adding a check that number of days cannot be less than 1 and that there is also a max, say one month (30 days)

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,129 @@
...
+   * @param startDate The start date for the trip. Must be non-null.
+   * @param tripDays The list of tripDays. Must be non-null.
+   */
+  public Trip(String tripName, String startDate, ArrayList<TripDay> tripDays) {
+    if (tripName == null) {
+      throw new IllegalArgumentException("tripName cannot be null");
+    }
+
+    if (startDate == null) {
+      throw new IllegalArgumentException("startDate cannot be null");
```
adamfweidman, line 77: I think is the one who is checking this but is it possible that the date won't be the right format? like a "65938123". That would be something we need to check against if users can input that

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    if (tripDays == null) {
+      throw new IllegalArgumentException("tripDays cannot be null. Use empty array instead.");
```
adamfweidman, line 81: Good error message!

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,129 @@
...
+
+    this.tripDays = new ArrayList<>();
+    this.tripDays.addAll(tripDays);
+  }
+
+  /**
+   * Returns the human-readable name for this trip.
+   */
+  public String getTripName() {
+    return tripName;
```
adamfweidman, line 99: Personal preference: I like to use this.tripName which just ensures that its the global variable. Your choice!

src/main/java/com/google/sps/Trip.java
zpChris: Building on this, this may be excess / unnecessary, but I'm wondering if you could add a check to determine whether the `numDays` value is consistent with the `startDate` and `endDate` variables. This may be difficult given that those are Strings though, so I'll leave this up to whatever you decide.

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * Trip is the class for storing a single trip (could be multiple days).
+ */
+public class Trip {
+  private String tripName;
+  private String startDate;
+  private String endDate;
+  private int numDays;
+  private ArrayList<TripDay> tripDays;
```
zpChris, line 27: nit: Personal preference from my CIS classes, I prefer a broader interface like `List<TripDay>` to be used here. However, I couldn't find anything in the style guide supporting this, so any action is up to you.

zpChris: (This same nit is present in different places, namely the parameters of different functions, but I don't want to saturate the review with nits so I'll just leave a single nit here.)

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,129 @@
...
+    if (tripDays == null) {
+      throw new IllegalArgumentException("tripDays cannot be null. Use empty array instead.");
+    }
+
+    this.tripName = tripName;
+    this.startDate = startDate;
+    this.endDate = endDate;
+    this.numDays = numDays;
+    this.tripDays = new ArrayList<>();
+    this.tripDays.addAll(tripDays);
```
zpChris, line 60: Is there a reason you don't just set `this.tripDays = tripDays`? Are you worried about accidentally modifying the original parameter, since it doesn't make a copy?

src/main/java/com/google/sps/Trip.java
zpChris, line 27: No action required, just a note. Eventually, I think we'll add a `String tripId` variable that holds a 16-digit alphanumeric identifier, so that we can easily store and access different trips in datastore. Again, no action required on this, just a note for later plans.

src/main/java/com/google/sps/Trip.java
zpChris: That's true, I think there are ways to do this! You could [check if the String is a valid date](https://stackoverflow.com/questions/33968333/how-to-check-if-a-string-is-date), and if it is, [convert it to Date format](https://www.javatpoint.com/java-string-to-date) in order to confirm the `startDate` and `endDate` values match the value of `numDays` and the size of `tripDays`.

src/main/java/com/google/sps/TripDay.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.ArrayList;
+
+/**
+ * TripDay objects store the trip/route information for a single day.
+ */
+public class TripDay {
+  private String origin;
+  private String destination;
+  private ArrayList<String> locations;
```
zpChris, line 25: Two questions:
1. Will these variables always store Strings, or are you going to later use the Google Maps Java Client to store place objects here? (I just saw in the `TripDayTest` class that these Strings are seemingly **place IDs** -- if this is the case, can you add a comment somewhere explaining this?)
2. Say the last location (POI) is also where the user wants to end their day. Is it included in both `locations` and as the `destination`, or is it no longer included at the end of `locations`?

zpChris: Expanding on this; on the frontend, I haven't provided users any options for putting a `destination` different than their `origin`, because at least for the MVP I assumed this would be the same. Was this a fair assumption, or should I include a spot for where they end their day as well?

My main concern here is that this could become overly complicated once we start to add multiple days -- I'd have to dynamically allow adding different `origin` and `destination` values for every day.

src/test/java/com/google/sps/TripTest.java
```diff
@@ -0,0 +1,94 @@
...
+  @Test(expected = IllegalArgumentException.class)
+  public void testTripConstructorMultiDayNull() {
+    Trip trip = new Trip(TRIP_NAME, START_DATE, END_DATE, NUM_DAYS, null);
+  }
+
+  // Test single day constructor with null parameter
+  @Test(expected = IllegalArgumentException.class)
+  public void testTripConstructorSingleDayNull() {
+    Trip trip = new Trip(TRIP_NAME, START_DATE, null);
+  }
```
zpChris, line 92: If you end up adding code to verify that the dates are valid and that they match the other inputs, can we also add tests checking this functionality as well? (Only applicable if you end up adding that initial code, of course.)

src/main/java/com/google/sps/TripDay.java
eshika: 1. I have these as Strings (for Place IDs) for now, but this could change depending on how I end up using the Java Client! Still exploring/learning about the Client so will keep you posted on that. Have added a comment explaining that these params are Place ID Strings for now!

2. If the last POI is the last stop on the route, then that will be stored only in destination, not locations. For the MVP, I am just assuming that the origin and destination are the same, so this should not be an issue.

For the MVP, I think the form is fine with just having the user input an origin because we are assuming that all trips will be round trips to and from that location, so I can just set the destination to the origin when calling the DirectionsQuery. Beyond the MVP, we can talk about how we want it to look, but I don't think there is any immediate action required for now.

src/main/java/com/google/sps/Trip.java
eshika: Yes! That was the reasoning for this, added a comment in the code to clarify :)

src/main/java/com/google/sps/Trip.java
eshika: Added a check to ensure the number of days is between 1 and 31 (inclusive), and am now computing the number of days using the SimpleDateFormat library so that it is consistent with the startDate and endDate.

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,190 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Date;
```
bheff-work, line 19: Since Java 8 came out, the preference is to use the java.time libraries, not java.util for working with time. Oracle themselves recommend this (https://docs.oracle.com/javase/tutorial/datetime/iso/legacy.html).

Let me know if you have any questions about it!

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,190 @@
...
+
+  /**
+   * Creates a new Trip that is only one day by default. 
+   * This constructor will be used for the MVP.
+   * 
+   * @param tripName The human-readable name for the trip. Must be non-null.
+   * @param startDate The start date for the trip. Must be non-null.
+   * @param tripDays The list of tripDays. Must be non-null.
+   */
+  public Trip(String tripName, String startDate, List<TripDay> tripDays) {
```
bheff-work, line 93: I believe you can call your other constructor from here, which will reduce a lot of the code duplication.

public Trip(String tripName, String startDate, List<TripDay> tripDays) {
  this(tripName, startDate, startDate, tripDays);
}

src/main/java/com/google/sps/TripDay.java
```diff
@@ -0,0 +1,77 @@
...
+   */
+  public String getDestination() {
+    return this.destination;
+  }
+
+  /**
+   * Returns an List<String> of locations for this TripDay.
+   */
+  public List<String> getLocations() {
+    return this.locations;
```
bheff-work, line 75: same comment as in the Trip class, you'll want to make a copy of this list if you don't want it to be modified

src/test/java/com/google/sps/TripDayTest.java
```diff
@@ -0,0 +1,56 @@
...
+    Assert.assertEquals(tripDay.getOrigin(), HOTEL_ID);
+    Assert.assertEquals(tripDay.getDestination(), HOTEL_ID);
+    Assert.assertEquals(tripDay.getLocations(), locations);
+  }
+
+  // Test TripDay constructor with null parameter
+  @Test(expected = IllegalArgumentException.class)
+  public void testTripDayConstructorNull() {
+    TripDay tripDay = new TripDay(HOTEL_ID, HOTEL_ID, null);
+  }
```
bheff-work, line 55: It would be a good idea to test all the expected failures

src/test/java/com/google/sps/TripTest.java
```diff
@@ -0,0 +1,125 @@
...
+package com.google.sps;
+
+import com.google.sps.Trip;
+import com.google.sps.TripDay;
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import java.util.ArrayList;
+import java.util.Arrays;
```
bheff-work, line 24: nit: order your imports alphabetically

src/test/java/com/google/sps/TripTest.java
```diff
@@ -0,0 +1,125 @@
...
+    
+    tripDays = new ArrayList<>();
+    tripDays.add(tripDay);
+
+    Trip trip = new Trip(TRIP_NAME, "06-28-2020", END_DATE, tripDays);
+  }
+
+  // Test multiple day constructor with over a month
+  @Test(expected = IllegalArgumentException.class)
+  public void testTripConstructorMultiDayLongTrip() {
```
bheff-work, line 100: this is great, since it captures the edge case of being at 31 days, but you also need a test for the edge case of it's '0'. That test would have caught your off-by-one bug in Trip. Whenever you have restrictions like that in your code, it's important to test right on the edge of those restrictions

src/test/java/com/google/sps/TripTest.java
```diff
@@ -0,0 +1,125 @@
...
+    tripDays.add(tripDay);
+
+    Trip trip = new Trip(TRIP_NAME, START_DATE, tripDays);
+
+    Assert.assertEquals(trip.getTripName(), TRIP_NAME);
+    Assert.assertEquals(trip.getStartDate(), START_DATE);
+    Assert.assertEquals(trip.getEndDate(), START_DATE);
+    Assert.assertEquals(trip.getNumDays(), 1);
+    Assert.assertEquals(trip.getTripDays(), tripDays);
+
```
bheff-work, line 78: nit: remove extra line

src/test/java/com/google/sps/TripTest.java
```diff
@@ -0,0 +1,125 @@
...
+  // Test multiple day constructor with wrong date format
+  @Test(expected = IllegalArgumentException.class)
+  public void testTripConstructorMultiDayInvalidDate() {
+    TripDay tripDay = new TripDay(HOTEL_ID, HOTEL_ID, locations);
+    
+    tripDays = new ArrayList<>();
+    tripDays.add(tripDay);
+
+    Trip trip = new Trip(TRIP_NAME, "06-28-2020", END_DATE, tripDays);
+  }
```
bheff-work, line 96: test invalid end date too

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,190 @@
...
+import java.text.SimpleDateFormat;
+
+/**
+ * Trip is the class for storing a single trip (could be multiple days).
+ */
+public class Trip {
+  private static final SimpleDateFormat DATE_FORMAT = new SimpleDateFormat("yyyy-MM-dd");
+
+  private String tripName;
+  private String startDate;
```
bheff-work, line 30: Perhaps it would be best to keep the date fields in an actual time-based object? That way people wouldn't have to convert it themselves and it makes it easier to display however they would like?

eshika:
Few notes:
1. Per Adam's comment, I have added a constant 31 day trip limit to multi day trips, but this can be increased/modified later as needed.
2. Per Bootsie's comment, I have switched to using the LocalDate library for storing Trip StartDate and EndDate and calculating the number of days in a trip.

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Creates a new Trip.
+   *
+   * @param tripName The human-readable name for the trip. Must be non-null.
+   * @param startDate The start date for the trip. Must be non-null. Must be in yyyy-MM-dd date format.
+   * @param endDate The end date for the trip. Must be non-null. Must be in yyyy-MM-dd date format.
+   * @param tripDays The list of tripDays. Must be non-null.
+   */
+  public Trip(String tripName, String startDate, String endDate, List<TripDay> tripDays) {
```
bheff-work, line 45: Why not take in LocalDateTime objects as startDate and endDate? Or at least have another constructor that does?

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   * Returns the number of days in this trip.
+   */
+  public int getNumDays() {
+    return this.numDays;
+  }
+
+  /**
+   * Returns an List<TripDay> copy of TripDays for this trip.
```
bheff-work, line 126: nit: replace 'an' with 'a'

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,156 @@
...
+
+    if (tripDays == null) {
+      throw new IllegalArgumentException("tripDays cannot be null. Use empty array instead.");
+    }
+
+    if (!isValidDate(startDate)) {
+      throw new IllegalArgumentException("Invalid startDate format. Must be in yyyy-MM-dd date format.");
+    }
+
+    if (!isValidDate(endDate)) {
```
bheff-work, line 66: not sure you need these methods anymore, since you're doing the LocalDate.parse() below. Maybe have the method check *and* create the LocalDate? That way you aren't having to parse twice, and you don't have to have two try catches in the constructor

src/main/java/com/google/sps/Trip.java
```diff
@@ -0,0 +1,156 @@
...
+    return true;
+  }
+
+  /**
+   * Calculates number of days between two dates.
+   * @param startDate The LocalDate representation of start date.
+   * @param endDate The LocalDate representation of end date.
+   */
+  private static int calcNumDays(LocalDate startDate, LocalDate endDate) {
+	  int numDays = ((int) ChronoUnit.DAYS.between(startDate, endDate)) + 1;
```
bheff-work, line 153: indentation is off here

src/test/java/com/google/sps/TripTest.java
bheff-work: I still don't see a test that checks for trip of length 0? Am I missing it?

src/main/java/com/google/sps/Trip.java
eshika: The user input form gives us the dates as Strings. Since we'll be using the user input to construct the Trips, I chose to design the constructor this way.

src/test/java/com/google/sps/TripTest.java
eshika: Good catch! Because a trip with the same start and end date is considered to have 1 day, a trip of length 0 would occur when the end date is one day before the start date. I added a test for this to ensure that the IllegalArgumentException is thrown.

PR #18, opened 2020-07-01
Add user authentication, update login / logout links, update and add header to index, maps, and calendar page.

zpChris:
List of changes:
1. Add logic in `AuthServlet.java` to detect whether the user is signed in or not.
2. Update the "sign in" and "sign out" buttons on the frontend (POST requests were placed with actual links).
3. Update header on index and maps pages.
4. Add the header to the `calendar.html` page.
5. Add jQuery to calendar HTML and JS pages.
6. Extract the authorization code into a separate file, `authScript.js`. This file also handles setting up the basic structure of the site (which currently only includes the header).

Images of the changes:
Local sign in page (on deployment, this changes to an official Google sign-in page -- the local version is auto-generated just for testing):
image
Calendar page:
image

The other pages with the header have a revised link for the "Sign out" button, but the design is the same.

In the future:
- Currently, the header is copy / pasted in each HTML file (`index.html`, `maps.html`, `calendar.html`, and `index.html` in the `/trips` folder); since this is duplicate code, we'd eventually want to remove this and add it to a separate function. However, since I haven't gotten a response from the VSEs regarding handlebars.js, I am moving forward with this implementation for the moment.
- The `authScript.js` file will eventually also construct the basic structure of the site, but it currently only adds the header. Soon, we'll want to set a width for the site and other characteristics that make the design uniform across the different pages.


src/main/webapp/calendar-script.js
```diff
@@ -12,8 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
-document.addEventListener('DOMContentLoaded', function() {
+// Triggered upon DOM load.
+$(document).ready(() => {
```
zpChris, line 16: This is only personal preference, as I used jQuery elsewhere for this use. I'm happy to change this back if wanted.

zpChris:
Explaining some of the relevant changes:
- The "Sign out" button in the header was changed to an `a` element inside of a `div` because there no longer needs to be a POST request in order to sign out, it is just a link. This link does not immediately have a `href` attribute because this link is created from the backend in the `AuthServlet.java` file.
- The header was added across all HTML files.
- The `isSignedIn()` function in `script.js` was replaced by `getAuthObject()`, as this allows you to get the user's sign-in status _as well as_ get other relevant sign-in information, including the `href` link for the "sign in" and "sign out" buttons.

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -23,28 +26,86 @@
...
-    signedIn = false;
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("application/json");
+
+    // Set up user auth objects.
+    UserService userService = UserServiceFactory.getUserService();
+    UserAuth userAuth;
+
+    // Set redirect URL after login / logout as index page.
+    final String redirectUrl = "/";
```
adamfweidman, line 38: add: static

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -23,28 +26,86 @@
...
 
+  /**
+   * Inner class that holds relevant login/logout and user information.
+   */
+  class UserAuth {
+    // Fields that hold relevant login data.
+    private boolean loggedIn;
+    private String loginUrl;
+    private String logoutUrl;
+    private String email;
```
adamfweidman, line 89: I made a similar class in my portfolio and bootsie told me that:
- the loggedIn boolean is not necessary since we can always check whether the email is null or not
- correct me if im wrong. We are only ever going to need to have one link: either login or logout so we don't need 2 url links.
I'd me happy to show you my class if you wanted to take a look.

src/main/webapp/calendar-script.js
adamfweidman: as long as it doesn't change functionality (it shouldn't) then I'm good with this. I dont have a preference

src/main/webapp/trips/tripScript.js
```diff
@@ -12,36 +12,33 @@
...
       setContentWidth('800px');
-      displayHeader();
     }
   });
 });
 
-// Returns Promise with the sign in status in a boolean.
-function isSignedIn() {
+// Returns Promise with the auth object, containing login status and information.
+function getAuthObject() {
```
adamfweidman, line 32: this function is also in authScript.js. Is it necessary in both... (genuine question)

src/test/java/com/google/sps/data/User.java
```diff
@@ -0,0 +1,21 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.data;
+
+/**
+ * The User class holds all the relevant account and identification information
+ * of the user.
+ */
+public class User {}
```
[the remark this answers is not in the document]
zpChris: I think you were looking at an old version of the code, I removed this class (and thus there shouldn't be any missing ending line).

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: Yeah we should talk tomorrow -- I have reasons for how I structured the class this way, though I'm open to changing it.

src/main/webapp/trips/tripScript.js
zpChris: That's a good point, I actually don't need this -- thanks! Removed :)

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: Talked with Bootsie, and I took this advice!

src/main/webapp/script.js
```diff
@@ -21,11 +21,19 @@ $(document).ready(() => {
   // Set up trigger to add hidden POI form elements upon submission.
   addHiddenPoiFormTrigger();
 
-  isSignedIn().then((signInStatus) => {
+  getAuthObject().then((authObject) => {
     // Display the sign-in page or "start trip" form depending on sign in status.
-    if (signInStatus) {
+    if (authObject.email !== null) {
```
tristan-jh, line 26: After logging out I still see the trip creation page. I think the 'hasOwnProperty' or 'in' operators would do what you want instead of !== null.  Checking for null/non-existence in javascript is fiddly :/

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -23,28 +26,76 @@
...
-    signedIn = false;
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("application/json");
+
+    // Set up user auth objects.
+    UserService userService = UserServiceFactory.getUserService();
+    UserAuth userAuth;
+
+    // Set redirect URL after login / logout as index page.
+    final String redirectUrl = "/";
```
tristan-jh, line 38: optional: I'll second Adams comment I think this could reasonably be a class constant since it's always the same regardless of the context in which this method was called.

src/main/webapp/script.js
zpChris: `hasOwnProperty` returns `true` if the property is present, but set to `null`; I didn't see anything on using `in` to check for `null` values in JavaScript? I removed the `loggedIn` boolean property because Bootsie and Adam argued it was unnecessary, as I could check for `null` here -- what would you advise I do?

zpChris: Actually, `hasOwnProperty` works! I didn't realize that the `email` object isn't present; in the backend, when the property is set to `null` and the object is converted to JSON through GSON, the property disappears. I just updated this, I think it should work now -- thanks for verifying this.

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: Done!

zpChris: Actually, I still have to add static; changing this now.

zpChris: Done!

PR #21, opened 2020-07-08
Change Event Class to use Time lib

adamfweidman:
the files changed is confusing to read but essentially I simplified the class:
1. I used LocalDateTime library to keep track of my date and time 
2. I deleted all my functions that created the time strings with the proper format, funciton to add certain amount of time etc, (this can all be done in the LocalDateTime library.
3. I changed the tests to reflect this and additionally I tested my `getProperDateFormat()` function as I need to be able to output correct string in order to put it into fullCalendar library on the frontend.


src/main/java/com/google/sps/data/Event.java
```diff
@@ -13,9 +13,11 @@
 // limitations under the License.
  
 package com.google.sps.data;
- 
-import java.text.SimpleDateFormat;
-import java.util.Date;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
```
zpChris, line 19: Order imports alphabetically, so this goes at the top.

src/main/java/com/google/sps/data/Event.java
```diff
@@ -48,119 +45,54 @@
...
-   * @param date date from user input travel day. Cannot be null.
-   * @param startTime start of activity (HHMM format)
+   * @param startTime start of activity (YYYY-MM-DD'T'HH:MM:SS)
    * @param travelTime time spent traveling to next location (minutes). 
    *        Null if last location of the day.
    * @param timeAtLocation time spent at POI (minutes)
    */
-  public Event(String name, String address, String date, int startTime, 
+  public Event(String name, String address, LocalDateTime startTime, 
               int travelTime, int timeAtLocation) {
```
zpChris, line 54: Is there a reason why `timeAtLocation` is an `int`, which is then converted to a `long`? Could it simply initially be a `long`? (That being said, it's a simple switch, so no big deal if you want to keep this structure.)

src/main/java/com/google/sps/data/Event.java
```diff
@@ -48,119 +45,54 @@
...
-      throw new NullPointerException("Date cannot be null");
-    }
-
+  private static void checkTravelTime(long time) {
     if (time < MIN_POSSIBLE_TIME) {
       throw new IllegalArgumentException("Time cannot be less than " + MIN_POSSIBLE_TIME);
     }
 
     if (time >= MINUTES_IN_A_DAY) {
       throw new IllegalArgumentException("Time cannot be more than " + MINUTES_IN_A_DAY);
```
zpChris, line 88: I see you didn't changed than, but should the error be "Time cannot be more than **or equal to** " + MINUTES_IN_A_DAY?

src/main/java/com/google/sps/data/Event.java
adamfweidman: I put it as an int that way theres no error when creating the event since I will probably get an int from Eshika (Long to int conversions sometimes throws errors).

src/main/java/com/google/sps/data/Event.java
```diff
@@ -48,119 +46,54 @@
...
-    createCalendarTimes();
+    this.endTime = startTime.plusMinutes(Long.valueOf(timeAtLocation));
+    this.travelTime = Long.valueOf(travelTime);
+    checkTravelTime(this.travelTime);
   }
- 
+
   /**
-   * Constructor that assume one hour default spent at location
+   * Constructor that takes in time spent at location
```
bheff-work, line 65: I think this may have been a copy-paste error, you probably still want "Constructor that assume one hour default spent at location" as the comment here

src/main/java/com/google/sps/data/Event.java
```diff
@@ -48,119 +46,54 @@
...
-      numHours = "0" + numHours;
-    } 
-
-    if (Integer.parseInt(numMins) < 10) {
-      numMins = "0" + numMins;
-    }
-
-    return numHours + numMins;
+  public static String getProperDateFormat(LocalDateTime ldt) {
+    DateTimeFormatter dtf = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss");
```
bheff-work, line 94: It looks like you're using the ISO_LOCAL_DATE_TIME format, if that's the case, you can just write:

"return DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(ldt);"

And if this is the case, then I'm not sure how necessary this method actually is. It's referring to a 'proper' date format, but what does that actually mean? A user of your event class could want the time in a multitude of formats.

src/test/java/com/google/sps/EventTest.java
```diff
@@ -15,6 +15,10 @@
 package com.google.sps;
 
 import com.google.sps.data.Event;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.LocalTime;
```
bheff-work, line 21: you'll want to move 'java.time.LocalTime' up one line

src/test/java/com/google/sps/EventTest.java
```diff
@@ -35,90 +39,58 @@
...
-  public void testMinuteToHHMMFormatThreeHour() {
-    String output = Event.convertToFormat(THREE_HOURS);
-    Assert.assertEquals("0300", output);
-  }
+  private static final LocalDateTime DEF_LDT = LocalDateTime.of(LocalDate.parse("2020-06-25"), LocalTime.of(10, 0));
 
   @Test(expected = IllegalArgumentException.class)
-  public void testMinuteToHHMMFormatLessThanZero() {
-    Event.convertToFormat(-1);
+  public void testcheckTravelTimeMinPossibleTime() {
```
bheff-work, line 46: nit: I'd rename this test to make it clearer that you expect a failure, e.g. testTravelTimeBelowMinPossible

src/test/java/com/google/sps/EventTest.java
```diff
@@ -35,90 +39,58 @@
...
+    Event e = new Event (GOLDEN_GATE_PARK, ADDRESS, DEF_LDT, -1, HOUR);
   }
 
   @Test(expected = IllegalArgumentException.class)
-  public void testMinuteToHHMMFormatGreaterThanFullDay() {
-    Event.convertToFormat(1440);
+  public void testcheckTravelTimeGreaterThanFullDay() {
```
bheff-work, line 51: you've forgotten to camelCase 'Check' on a lot of these tests

src/main/java/com/google/sps/data/Event.java
```diff
@@ -48,119 +46,54 @@
...
-   * @param date date from user input travel day. Cannot be null.
-   * @param time time to to start or end event. Must be within bounds
+   * @param time amount of time spent traveling (minutes)
    */
-  private static String createStrTime(String date, int time) throws Exception {
-    if (date == null) {
-      throw new NullPointerException("Date cannot be null");
-    }
-
+  private static void checkTravelTime(long time) {
```
bheff-work, line 83: not sure this method is necessary, but I'm not overly fussed if you keep it

src/main/java/com/google/sps/data/Event.java
```diff
@@ -48,119 +46,54 @@
...
     this.name = name;
     this.address = address;
-    this.date = date;
     this.startTime = startTime;
-    this.travelTime = travelTime;
-    this.endTime = calculateEndTime(timeAtLocation);
-  
-    createCalendarTimes();
+    this.endTime = startTime.plusMinutes(Long.valueOf(timeAtLocation));
+    this.travelTime = Long.valueOf(travelTime);
```
bheff-work, line 60: why does travelTime need to be converted from an int to a long? If you're always going to get an 'int' why change it? It doesn't seem to be used as a long anywhere.

adamfweidman: In the LocalDateTime library the plusMinutes takes in a long not an int

PR #24, opened 2020-07-09
Create events dynamically from user input

adamfweidman:
This is the main part of my MVP calendar:
1. Added start and end times in (yyyy-MM-dd'T'HH:mm:ss) format to the Event class to be used in  calendar-script.js 
2. Calendar script: iterate through all the events on the /calculate-trip url and dynamically add them to the calendar
3. eventDoPost(): go through the inputs from the user (gotten from Chris) and search for parameter that have "poi" in them. For reference all the POIs are listed as "poi-NUMBER_ADDED". For all those poi's I then create an event and store it in datastore.
4. eventDoGet(): Here I simply get all events in datastore and write them to the URL so that the calendar can read it later.

p.s. for the moment I do not have access to different driving times and amount of time spent at each location. Therefore, I manually move the start time of every event by 90 mins (60mins at location and 30mins driving). (This is the purpose of the setDateTime() function). This will be removed in versions going forward.

### DIFF
--- a/src/main/java/com/google/sps/servlets/TripServlet.java
+++ b/src/main/java/com/google/sps/servlets/TripServlet.java
@@ -71,7 +71,8 @@ public class TripServlet extends HttpServlet {
   }
 
   @Override
-  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+  public void doPost(HttpServletRequest request, HttpServletResponse response) 
+      throws IOException {
     response.setContentType("application/json;");
 
     // do post for events
@@ -83,7 +84,8 @@ public class TripServlet extends HttpServlet {
    * Make the servlet cleaner
    * Iterate through the entities and create the events and write them to json
    */
-   private void eventDoGet(HttpServletRequest request, HttpServletResponse response) throws IOException { 
+  private void eventDoGet(HttpServletRequest request, HttpServletResponse response)
+      throws IOException { 
     Query query = new Query("events");
 
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
@@ -107,10 +109,12 @@ public class TripServlet extends HttpServlet {
    }
 
   /**
-   * Make the servlet cleaner
-   * Searches through the parameters and creates the events and puts them into datastore
+   * Make the servlet cleaner.
+   * Searches through the parameters and creates the events and puts them into
+   * datastore.
    */
-  private void eventDoPost(HttpServletRequest request, HttpServletResponse response) throws IOException { 
+  private void eventDoPost(HttpServletRequest request, HttpServletResponse response) 
+      throws IOException { 
     // Print out params to site to verify retrieval of "start trip" user input.
     Enumeration<String> params = request.getParameterNames();
 
@@ -138,11 +142,12 @@ public class TripServlet extends HttpServlet {
         Entity eventEntity = event.buildEventEntity();
 
         // put entity in datastore
-        DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+        DatastoreService datastore = 
+                                DatastoreServiceFactory.getDatastoreService();      
         datastore.put(eventEntity);
 
         // sets start time for next event 2 hours after start of prev
-        startDateTime = startDateTime.plusMinutes(Long.valueOf(NINETY_MINS)); 
+        startDateTime = startDateTime.plusMinutes(Long.valueOf(NINETY_MINS));
       }
 
       // redirect to home page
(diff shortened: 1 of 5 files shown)


src/main/webapp/trips/tripScript.js
```diff
@@ -23,7 +23,23 @@ $(document).ready(() => {
...
+    });
+  });
+}
+
+function displayHeader() {
+  // Display header for site.
+  document.getElementById('header').style.display = 'block';
+}
+
+// Set the width of the content container
```
zpChris, line 42: I'm

src/main/webapp/trips/tripScript.js
```diff
@@ -23,7 +23,23 @@ $(document).ready(() => {
...
+      reject(error);
+    });
+  });
+}
+
+function displayHeader() {
+  // Display header for site.
+  document.getElementById('header').style.display = 'block';
+}
+
```
zpChris, line 41: I'm assuming these lines were added during some sort of merge issue, as I think this was the code I wrote; it isn't necessary any more, so you can remove it. (You can tell because it isn't called anywhere.)

src/main/webapp/calendar-script.js
```diff
@@ -22,45 +22,25 @@ $(document).ready(() => {
...
   });
+  getEvents(calendar);
   calendar.render();
 });
+
+/**
+ * retrives the events from /calculate-trip url and dynamically adds the events
+ */
+function getEvents(calendar) {
+  fetch('/calculate-trip').then(response => response.json()).then((events) => {
```
zpChris, line 36: This is fine right now, just adding because I have a question about future direction: should we consolidate the GET function for Maps, Calendar, and the future Edit page under `/calculate-trip`, or should we have separate servlets for getting the relevant information and formatting it for all of these pages (i.e.; maps fetches `/get-maps`, calendar fetches `/get-calendar`, etc)?

adamfweidman: I think probably separate servlets!

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -8,26 +14,147 @@
...
   }
 
+  /**
+   * Converts list of Event objects into a JSON string using the Gson library.
+   */
+  private String convertToJson(List<Event> events) {
+    Gson gson = new Gson();
+    String json = gson.toJson(events);
+    return json;
+  }
```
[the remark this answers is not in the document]
adamfweidman: I had some trouble converting From Object to Event and back! For the moment I'm going to leave it as is but I can look into it more later.

adamfweidman:
In response to creating a separate servlet for the doGet methods. I will do this in a later CL as I've made the code more modular in a further branch and it is easy to separate.

src/main/webapp/calendar-script.js
adamfweidman: I've done this in my later CL where I also Post and Get TripDays

adamfweidman:
Yes Chris you are correct. I've done that in a separate CL where additionally I create TripDay entities. That CL is ready but I will wait to send it out until this is approved!

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -8,26 +14,146 @@
...
+/** 
+ * Servlet that currently allows for retrieving info from url and put in 
+ *  calendar
+ */
 @WebServlet("/calculate-trip")
 public class TripServlet extends HttpServlet {
 
+  // temporary vars
+  private LocalDateTime startDateTime;
+  private static int count = 0;
```
tristan-jh, line 46: this field is unused.

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -8,26 +14,146 @@
...
     while (params.hasMoreElements()) {
-      String paramName = params.nextElement();
-      response.getWriter().println(paramName + ": " + request.getParameter(paramName));
+      String p = params.nextElement();
+
+      /** 
+       * for each poi create the necessary fields, this will change
+       * as we are able to pull from the maps backend api
+       */
+      if (p.contains("poi")) {
```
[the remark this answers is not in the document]
adamfweidman: Sounds good!

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -8,26 +14,145 @@
...
     response.setContentType("application/json;");
+    
+    // do get for events
+    eventDoGet(response);
+  }
+
+  // function to set time, need to do only once
+  public void setDateTime(String date) {
+    if (startDateTime == null) {
+      startDateTime = LocalDateTime.of(LocalDate.parse(date), LocalTime.of(10, 0));
```
tristan-jh, line 69: I'm seeing a compilation error on this line since startDateTime is no longer a class field. I think this method should have a LocalDateTime return type instead of void

adamfweidman: Done! I didn't catch that

PR #25, opened 2020-07-10
Dynamically add trip cards to the "/trips" page. (On hold)

zpChris:
Leaving this CL, will re-reference it later once I finish other components.

No review currently needed.


zpChris:
I am removing this CL because these changes were incorporated into other CLs instead.

PR #27, opened 2020-07-13
Alter TripDay, put tripDays into datastore, create separate servlet

adamfweidman:
**Code Changes:**

1. I added a date field in TripDay as I feel it is necessary
2. I put a tripDay object into datastore based off the user input and then put the events into datastore (this is largely the same code from last CL).
3. I added a _date_ field to the TripDay class since as this project continues it will be necessary so the calendar knows which events should be placed on which dates.
4. In order to put the events in my calendar I need the events to be the only json objects at a url page. This was going to create problems as Eshika also needed a servlet to post the information that she would need. Therefore I created a new Servlet which gets all the events associated with their tripDay parent (built into the datastore hierarchy structure).

**Tests**
- TripServletTest: Here I test my doPost functionality (only had to mock a few methods). As per Tristan and Bootsie's comments instead of mocking data structure like entities or events/tripDays I create those and use those for testing.
- CalendarServletTest: I test the doGet method which simply checks that the right string is being printed to the .../get-calendar url which is used in the front end to fetch the events.

**Concerns:**
Eshika, Chris, and I all have a CL out that changes the TripServlet and TripServletTest files. Specifically, Eshika slightly altered the functionality of a method I wrote (this was beneficial! much appreciated) which I later refractured in this CL. Combining these two is going to take some effort to ensure that there are no merge conflicts. I wonder in what order we should be merging these CLs and how to avoid any code-breaking changes,


pom.xml
```diff
@@ -63,6 +63,12 @@
       <artifactId>commons-text</artifactId>
       <version>1.7</version>
     </dependency>
+    
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>2.0.5-beta</version>
+    </dependency>
```
zpChris, line 71: Note: I think @eshika used a different Mockito dependency. I don't think this is a pressing concern though, but I just wanted to note it down.

src/main/java/com/google/sps/data/Event.java
```diff
@@ -110,8 +111,8 @@ private static void checkTravelTime(long time) {
   /**
    * Build entity from event to be put in datastore off event attributes
    */
-  public Entity eventToEntity() {
-    Entity eventEntity = new Entity("events");
+  public Entity eventToEntity(Key parentKeyID) {
+    Entity eventEntity = new Entity("event", parentKeyID);
```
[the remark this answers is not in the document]
zpChris: Also, this is just a suggestion, you could add a test for this in `EventTest.java`, to make sure that the parent of the Entity matches the `parentKeyID`.

src/test/java/com/google/sps/TripDayTest.java
```diff
@@ -63,12 +66,24 @@ public void testTripDayConstructorNullDestination() {
...
+  // Test TripDay null date
+  @Test(expected = IllegalArgumentException.class)
+  public void testTripDayConstructorNullDate() {
+    TripDay tripDay = new TripDay(HOTEL_ID, HOTEL_ID, locations, null);
+  }
+
+  // Test TripDay empty date
+  @Test(expected = IllegalArgumentException.class)
+  public void testTripDayConstructorEmptyDate() {
+    TripDay tripDay = new TripDay(HOTEL_ID, HOTEL_ID, locations, "");
```
zpChris, line 87: You may want to add tests that check whether the methods that convert the TripDay to an Entity and back work.

src/test/java/com/google/sps/TripServletTest.java
```diff
@@ -0,0 +1,132 @@
...
+  private static final String INPUT_DATE = "2020-07-15";
+  private static final String POI_ONE = "one";
+  private static final String POI_TWO = "two";
+
+  // create TripServlet object
+  TripServlet tripServlet;
+
+  // initialize mock objects
+  HttpServletRequest request = mock(HttpServletRequest.class);
+  HttpServletResponse response = mock(HttpServletResponse.class);
```
zpChris, line 56: Should you set access levels on these field variables?

src/test/java/com/google/sps/TripServletTest.java
```diff
@@ -0,0 +1,132 @@
...
+    List<Entity> eventEntities = tripServlet.putEventsInDatastore(request, response, params, tripDayEntity, INPUT_DATE, datastore);
+    Query query = new Query("event");
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> listResults = results.asList(FetchOptions.Builder.withDefaults());
+
+    // check that size is correct, added in correct order, entities match
+    Assert.assertEquals(2, listResults.size());
+    Assert.assertEquals(listResults.get(0), eventEntities.get(0));
+    Assert.assertEquals(listResults.get(1), eventEntities.get(1));
+  }
```
zpChris, line 131: I would consider adding a TODO to add a separate integration test for the full `doPost(...)` function.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,80 @@
...
+    doGetEvents(response, datastore);
+  }
+
+  /**
+   * Gets the events and prints them to writer, necessary break up for testing
+   */
+  public void doGetEvents(HttpServletResponse response, 
+      DatastoreService datastore) throws IOException {
+    // will add logic so that the Trip is gotten first and then all the tripDays
+    Query tripDayQuery = new Query("trip-day");
```
zpChris, line 55: I would set `"trip-day"` as a constant, ideally in the TripDay class.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,80 @@
...
+    PreparedQuery tripDayResults = datastore.prepare(tripDayQuery);
+
+    List<Event> events = new ArrayList<>();
+
+    // iterate through all the TripDays and get events for each one
+    for (Entity tripDayEntity : tripDayResults.asIterable()) {
+      Query eventsQuery = new Query("event", tripDayEntity.getKey());
```
zpChris, line 62: I would set `"event"` as a constant, ideally in the Event class.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.gson.Gson;
+import com.google.sps.data.Event;
+import com.google.sps.TripDay;
+import java.io.IOException;
+import java.time.LocalTime;
```
zpChris, line 27: I don't think this import is used?

src/main/java/com/google/sps/TripDay.java
```diff
@@ -76,4 +101,48 @@ public String getDestination() {
     locationsCopy.addAll(this.locations);
     return locationsCopy;
   }
+
+  /**
+   * Add locations
+   */
+  public void addLocations(ArrayList<String> locations) {
+    this.locations = locations;
+  }
```
zpChris, line 110: Just to clarify, I meant the method `addLocation(...)` that would add the location one-by-one (and accept a `String`), but if this works as well then that should be fine.

adamfweidman: Eshika wil lbe adding the locations all together after she calculates the order so an `addLocation()` function isnt needed

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,78 @@
...
+public class CalendarServlet extends HttpServlet {
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) 
+      throws IOException {
+    response.setContentType("application/json;");
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // gets the events from datastore and writes them to .../get-calendar
+    doGetEvents(response, datastore);
```
zpChris, line 44: I'm not sure if a separate method is necessary here (if you created a new method to help with testing, you can still test `doGet(...)` directly; but it's mostly a personal preference, so no action required here either unless you want to change this.

src/test/java/com/google/sps/TripDayTest.java
```diff
@@ -63,12 +84,42 @@ public void testTripDayConstructorNullDestination() {
...
+  }
+
+   @Test
+  public void testbuildEntityAndFromEntity() {
```
zpChris, line 109: Fix camel case: test**B**uildEntityAndFromEntity.

src/main/java/com/google/sps/servlets/CalendarServlet.java
adamfweidman: I did this for tetsting so I could pass in a datastore object

src/main/java/com/google/sps/TripDay.java
zpChris: Ok gotcha, yeah that sounds good then!

src/main/java/com/google/sps/servlets/CalendarServlet.java
zpChris: AFAIK, you don't need to pass in a Datastore object; since you're not mocking anything, you can just call `doGet` directly.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,78 @@
...
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/get-calendar")
+public class CalendarServlet extends HttpServlet {
```
bheff-work, line 35: Add a class comment describing what this Servlet is for

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,78 @@
...
+    // gets the events from datastore and writes them to .../get-calendar
+    doGetEvents(response, datastore);
+  }
+
+  /**
+   * Gets the events and prints them to writer, necessary break up for testing
+   */
+  public void doGetEvents(HttpServletResponse response, 
+      DatastoreService datastore) throws IOException {
+    // will add logic so that the Trip is gotten first and then all the tripDays
```
bheff-work, line 52: If this is something you are planning to do in a later CL, I'd suggest putting the comment in this form:

// TODO(adamfweidman): Add logic so that the Trip is gotten first and then all the TripDays

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,78 @@
...
+    Query tripDayQuery = new Query(TripDay.QUERY_STRING);
```
bheff-work, line 53: Are these being restricted by user? I don't see where that is being done. We don't want to be sending every user's Trips back for every call (eek, privacy!).

If this is an interim step, and you're planning to add the User restriction later, I don't know if I'm okay with that. At the very least, I think we should hard-code a user id and throw in a big ol' TODO.

src/main/java/com/google/sps/servlets/CalendarServlet.java
bheff-work, line 35: This servlet seems to be just pulling TripDays, why is it called CalendarServlet? I feel like that could be useful for other parts of the system other than Calendar?

src/main/java/com/google/sps/TripDay.java
```diff
@@ -76,4 +101,48 @@ public String getDestination() {
     locationsCopy.addAll(this.locations);
     return locationsCopy;
   }
+
+  /**
+   * Add locations
+   */
+  public void addLocations(ArrayList<String> locations) {
+    this.locations = locations;
```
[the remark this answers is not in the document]
adamfweidman: The locations will be added just once after Eshika has calculated the correct order so a `setLocations()` is all that is needed

src/main/java/com/google/sps/TripDay.java
```diff
@@ -76,4 +101,48 @@ public String getDestination() {
...
+    Entity tripDayEntity = new Entity(QUERY_STRING, parentKeyID);
+    tripDayEntity.setProperty(ORIGIN, this.origin);
+    tripDayEntity.setProperty(DESTINATION, this.destination);
+    tripDayEntity.setProperty(DATE, this.date);
+    return tripDayEntity;
+  }
+
+  /**
+   * Build tripDay from entity. Not needed for this CL but will be necessary 
+   * later. NOTE: Unsure how to handle locations. Cannot store in datastore
```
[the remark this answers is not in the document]
adamfweidman: Yes yo are correct, another thought:

An issue that I've discussed with Eshika is that we cannot store List objects in datastore. We've decided to store the locations as children of the Trip Class in datastore. I'm unsure of how exactly the locations field will be used by Eshika but she specified that she needs it.

src/main/java/com/google/sps/TripDay.java
```diff
@@ -33,7 +44,7 @@
    * @param locations The list of POIs (list of Google Maps Place ID strings) 
                       that are stopovers for this day. Must be non-null.
    */
-  public TripDay(String origin, String destination, List<String> locations) {
+  public TripDay(String origin, String destination, List<String> locations, String date) {
```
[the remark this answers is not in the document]
adamfweidman: Note that I'm deleting the if statement in the constructor that checks null and empty string since the LocalDate library does this automatically

src/main/java/com/google/sps/servlets/CalendarServlet.java
adamfweidman: I don't know how Eshika is planning on puling information into the maps frontend but she doesn't use Events. I think she is planning on using a DirectionRequest object or something like that but if she can do that in this servlet as well than we can do that as well! The only problem is that in order to communicate with my calendar I need the events to be the only thing written by the `response.writer()`

src/main/java/com/google/sps/servlets/CalendarServlet.java
adamfweidman: Yes for the moment this is a privacy issue because all the Trips are being added to the page. I was waiting for Chris's CL to go through before I can add the code that restricts this behavior

src/main/java/com/google/sps/TripDay.java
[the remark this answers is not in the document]
adamfweidman: I'm a little confused what this means and how to solve it could we talk about it on Monday?

src/main/java/com/google/sps/servlets/CalendarServlet.java
adamfweidman: TODO added

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,78 @@
+// Copyright 2019 Google LLC
```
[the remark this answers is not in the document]
bheff-work: I believe this still needs to be addressed. Have you discussed this with Chris and Eshika, yet?

src/main/java/com/google/sps/servlets/CalendarServlet.java
bheff-work: Have you discussed with Eshika and Chris what we talked about this morning about having utility classes to do these repetitive calls for you?

adamfweidman:
I have discussed with Eshika and Chris and we are going to make a Utility class that will be used for the `doGet()` functionality of the Maps and Calendar Servlet.

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,112 @@
...
+  
+    Filter tripKeyFilter =
+      new FilterPredicate("__key__", FilterOperator.EQUAL, tripEntityKey);
+    Query tripQuery = new Query(Trip.TRIP, userEntity.getKey());
+    tripQuery.setFilter(tripKeyFilter);
+
+    PreparedQuery tripResults = datastore.prepare(tripQuery);
+    List<Entity> listResults = tripResults.asList(FetchOptions.Builder.withDefaults());
+
+     if (listResults.isEmpty()) {
```
bheff-work, line 80: nit: fix indentation

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,112 @@
...
+      throw new IllegalArgumentException("Trip does not exist");
+    }
+
+    // get the Entity object of the trip.
+    Entity tripEntity =  listResults.get(0);
```
bheff-work, line 85: If you are only expecting one result here, you should probably use the asSingleEntity method in PreparedQuery: https://cloud.google.com/appengine/docs/standard/java/javadoc/com/google/appengine/api/datastore/PreparedQuery.html#assingleentity

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,112 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) 
+      throws IOException {
+    response.setContentType("application/json;");
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    String stringTripKey = request.getParameter("tripKey");
+    Key tripKey = KeyFactory.stringToKey(stringTripKey);
+
+    // get current user
+    Entity userEntity = AuthServlet.getCurrentUserEntity();
```
bheff-work, line 60: What if there is no logged-in user?

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -156,44 +147,39 @@ public String getPlaceIdFromTextSearch(GeoApiContext context, String textSearch)
...
-
-    List<Event> events = new ArrayList<>();
-
-    // create the events
-    for (Entity entity : results.asIterable()) {
-      events.add(Event.eventFromEntity(entity));
-    }   
+  public Entity putTripDayInDatastore(HttpServletRequest request, 
+      DatastoreService datastore, LocalDate date, Key tripEntityKey) throws IOException {
+
```
bheff-work, line 159: nit: remove extra line at beginning of method

src/test/java/com/google/sps/TripServletTest.java
```diff
@@ -92,6 +110,67 @@ public void tearDown() {
   }
 
   @Test
+  public void testPutTripDayInDatastore() throws Exception {
+    // set mock object behavior
+    when(request.getParameter("inputDestination")).thenReturn(INPUT_DESTINATION);
+
+    // create key
+    long test = 123;
```
bheff-work, line 118: since you aren't using this variable again, I think it's fine to just write "KeyFactory.createKey("test", 123)"

src/test/java/com/google/sps/CalendarServletTest.java
```diff
@@ -0,0 +1,139 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import static org.mockito.Mockito.*;
```
bheff-work, line 41: nit: static imports generally go above non-static ones

src/test/java/com/google/sps/CalendarServletTest.java
```diff
@@ -0,0 +1,139 @@
...
+      "4265 24th Street San Francisco, CA, 94114";
+  private static final String INPUT_DATE = "2020-07-15";
+  private static final String TEST_NAME = "testName";
+  private static final String EMPIRE_ADDRESS = "20 W 34th St, New York, NY 10001";
+  private static final int HALF_HOUR = 30;
+  private static final LocalDateTime DEF_START_TIME = 
+      LocalDateTime.of(LocalDate.parse(INPUT_DATE), LocalTime.of(10, 0));
+  private static final String EMAIL = "test123@gmail.com";
+
+  // Cosntants to represent different Trip attributes.
```
bheff-work, line 59: nit: "Constants"

src/test/java/com/google/sps/CalendarServletTest.java
```diff
@@ -0,0 +1,139 @@
...
+    helper.setUp();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testWriteEventsCorrectly() throws Exception {
```
bheff-work, line 85: I think you're missing some test cases here e.g. no logged-in user, no trips returned, multiple trips returned, multiple events, multiple trip days

adamfweidman: Unless the user goes to .../calendar.html then there will always be a logged in user. there will always be a trip returned because once Chris's CL gets merged the URL will contain a tripID. I can add multiple events for a tripDay but for the MVP we are limiting the Trip to a single day

adamfweidman: So I looked into datastore and it does not allow you to give entities the same key so it is not possible for me to make that test. I did add a test that has multiple tripDays and events

src/main/java/com/google/sps/servlets/CalendarServlet.java
```diff
@@ -0,0 +1,118 @@
...
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) 
+      throws IOException {
+    response.setContentType("application/json;");
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // get current user
+    Entity userEntity = AuthServlet.getCurrentUserEntity();
+
+    System.err.println(request.getParameter("tripKey"));
```
bheff-work, line 59: remove println

PR #28, opened 2020-07-13
Add the Trip Entity from input form to Datastore, revise Trip class and tests, and add methods to AuthServlet.

zpChris:
### List of changes:
- Get form input from `/calculate-trip` POST request.
- Use the Google Maps Places Java Client API to get the `placeId` and **Place Details** of the `tripDestination`. This is used to get a **photo of the destination** and a **destination name**.
- Get the current logged-in User Entity from Datastore. Put the Trip Entity into Datastore, with the current logged-in User Entity as its ancestor.
- Revise the Trip class (and tests for the Trip class)  to hold necessary components to show "trip cards" on frontend.
- Add methods to the AuthServlet class to get the current logged-in User Entity, and revise the query method used to determine if the User is already in the database (by using a Query Filter).
- Redirect back to homepage if POST request is made with no user currently logged in.
- Add line in `.gitignore` to hide server-side `Config.java` file, which allows access to the Google Maps API in Java server-side files.

### Notes:
- In future CLs, lots of more code will be added to the `TripServlet.java` `doPost` method, so the different components will likely have to be further separated into methods.
- There is a possibility that the `tripKey` field of the Trip class should allowed to be `null`. This `tripKey` represents the **Key** of the Trip Entity in datastore, so if it is not yet in datastore, there should be no `tripKey`. However, I am operating with the assumption that the Trip should immediately be put in datastore before initialization, so I am leaving this unless there is an objection.
- I found that these methods may not be able to store POIs with non-English characters. This is interesting to note, but lower priority, so there's no action on this for now.
- The image source holds the API key (see the [Places API documentation](https://developers.google.com/places/web-service/photos#place_photo_requests) for why this is necessary).
- It is difficult to show that one Entity is the ancestor of another in Datastore just using the local UI. You can verify this by checking the key of each object; the key of the ancestor will be completely contained within the key of the child of the ancestor. You can also verify this in actual code; it is more difficult to show through testing, but this could be attempted.

### Screencast:
[Storing Trip data in Datastore](https://screencast.googleplex.com/cast/NTIyMjE0MzU4OTYxMzU2OHw1NGFkNDAyYy01YQ)

### Images:
Image of local datastore service storing the Trip after input form submission (the API key in the photo URL is replaced with **HIDDEN_API_KEY** by directly editing the HTML of the page -- if you tested, you would see an actual API key there):
image
Image of the photo URL in the above datastore example:
[Image of the Island of Hawai'i](https://lh3.googleusercontent.com/p/AF1QipM7tbCZOj_5SOft9cYgI7un3bmieieqvdYkCPT5=s1600-w400)


src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -17,13 +17,19 @@
...
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 import com.google.gson.Gson;
 import com.google.sps.data.User;
 import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
```
eshika, line 32: nit: order imports alphabetically

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -18,16 +33,117 @@
...
+    PlaceDetailsRequest placeDetailsRequest = PlacesApi.placeDetails(context, 
+      placeId);
+    try {
+      return placeDetailsRequest.await();
+    } catch(ApiException | InterruptedException e) {
+      throw new IOException(e);
+    }
+  }
+
+
```
eshika, line 137: nit: remove extra line

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: Thanks for the note, I missed that!

src/main/java/com/google/sps/servlets/TripServlet.java
zpChris: Done!

zpChris:
> The changes look good, but please add tests for the AuthServlet and TripServlet code. Also, in the future, it would be easier to review if you separate changes into smaller CLs (for example separate changes in the Trip class/tests from the changes in the servlets)

You bring up a good point @eshika; I'll separate these changes to separate CLs, because adding testing will make these files even larger (which I'm hoping to avoid).

PR #29, opened 2020-07-14
Dynamically add trips for each user to the "/trips/" page.

zpChris:
### List of changes:
- Create `UserTripServlet.java` file with a `doGet(...)` function to retrieve the list of trips for the current user.
- Add multiple functions (`getAndAddTripCards(...)`, `addTripCards(...)`, and `buildTripCard(...)`) to the `tripScript.js` file to build and display the trip cards.
- The **Calendar** and **Maps** buttons on the trip cards now hold a query of the form `?tripKey=TRIP_KEY_HERE` to the links, which will allow the Calendar and Maps pages to show the relevant trips information. For now, however, this query is ignored.
- Remove static / constant Trip cards from the `/trips/` page.
- Revise styling of Bootstrap "cards" by placing `width: 15rem` in the `style.css` file, maintaining uniformity across the site.

### Notes:
- There are currently no restrictions on the length of the trip name, which means the "card" design format could be poor if a user gave a very long title. We may want to add restrictions on the trip name length in the future, but for now, it's not a big priority.
- The images on the trip cards are not always the same size, which can create a mismatch in card height. This is again a difficult issue to fix, and not high priority, so I am ignoring this for now.
- I don't know how the trip cards are ordered. However, it doesn't seem to be random, so I am not placing an emphasis on sorting these a certain way.

### Screencast:
[Dynamically Adding Trips for Each User](https://screencast.googleplex.com/cast/NTE2MzczODk4MDE1NTM5MnwwOWUyZWVhZS00MA)

### Images:
Image of `/trips/` page with no added trips:
image

Image of `/trips/` page with Hawaii trip added:
image

Image of `/trips/` page with four trips:
image


zpChris:
I am closing this CL, as it will no longer be merged into CL #28, and I need to add sufficient testing. I will keep the branch open for reference however, and I will create new CLs based off this.

PR #31, opened 2020-07-15
Revise and add methods and tests for AuthServlet

zpChris:
### List of changes:
- Add methods to the AuthServlet class to get the current logged-in User Entity, and revise the query method used to determine if the User is already in the database (by using a Query Filter).
- Add tests for the AuthServlet class and its new methods.

### Notes:
- The name of this branch, **PutTripDatastore**, is not relevant anymore; I had planned to add code to the TripServlet class to put the Trip Entity objects in datastore, but I want to keep the CL a good size.

### Screencast:
No screencast is attached to this CL because the only change is to testing and the AuthServlet class.

### Images:
No images are attached to this CL because the only change is to testing and the AuthServlet class.

### DIFF
--- a/src/main/java/com/google/sps/servlets/AuthServlet.java
+++ b/src/main/java/com/google/sps/servlets/AuthServlet.java
@@ -64,7 +64,7 @@ public class AuthServlet extends HttpServlet {
       userAuth = new UserAuth(logoutUrl, userEmail);
 
       // Add user to database (if not already present).
-      addUserToDatabase(userEmail);
+      getOrCreateUserInDatabase(userEmail);
     } else {
       String loginUrl = userService.createLoginURL(redirectUrl);
 
@@ -91,7 +91,7 @@ public class AuthServlet extends HttpServlet {
    * This method returns the Entity object in the database (or the newly-created
    * user Entity, if none previously existed).
    */
-  public static Entity addUserToDatabase(String email) {
+  public static Entity getOrCreateUserInDatabase(String email) {
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
 
     // Only add User to database if they are not already present there.
@@ -126,12 +126,8 @@ public class AuthServlet extends HttpServlet {
     // Get user email from UserService.
     String email = userService.getCurrentUser().getEmail();
 
-    // Get user from datastore; if not present, add the user and return that.
-    Entity userEntity = getUserEntityFromEmail(email);
-    if (userEntity == null) {
-      Entity newUserEntity = addUserToDatabase(email);
-      return newUserEntity;
-    }
+    // Return user from datastore; if not present, add the user and return that.
+    Entity userEntity = getOrCreateUserInDatabase(email);
     return userEntity;
   }
 
@@ -147,7 +143,7 @@ public class AuthServlet extends HttpServlet {
     Query query = new Query(User.USER).setFilter(emailFilter);
     PreparedQuery results = datastore.prepare(query);
 
-    // Get the list of results. If empty, return null; otherwise, return true.
+    // Get the list of results.
     List<Entity> listResults = results.asList(FetchOptions.Builder.withDefaults());
     if (listResults.isEmpty()) {
       return null;
--- a/src/main/java/com/google/sps/servlets/AuthServlet.java
+++ b/src/main/java/com/google/sps/servlets/AuthServlet.java
@@ -17,13 +17,19 @@ package com.google.sps.servlets;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
 import com.google.appengine.api.users.UserService;
 import com.google.appengine.api.users.UserServiceFactory;
 import com.google.gson.Gson;
 import com.google.sps.data.User;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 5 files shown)


src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,80 @@ private static String convertToJson(UserAuth userAuth) {
...
+
+    // Return the user already in the database.
+    return userEntity;
+  }
+
+  /**
+   * Creates a new UserService object.
+   */
+  public static UserService getUserService() {
+    return UserServiceFactory.getUserService();
```
adamfweidman, line 117: I'm confused as to why this is a method and not just a regular function call. Is there a specific reason why breaking it up was necessary?

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,80 @@ private static String convertToJson(UserAuth userAuth) {
...
-  public void addUserToDatabase(String email) {
+  public static Entity addUserToDatabase(String email) {
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+
+    // Only add User to database if they are not already present there.
+    Entity userEntity = getUserEntityFromEmail(email);
+    if (userEntity == null) {
+      // User is not in database, create and add user object.
+      User newUser = new User(email);
+      Entity newUserEntity = newUser.buildEntity();
```
adamfweidman, line 102: unsure but do u test buildEntity in your User class?

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,80 @@ private static String convertToJson(UserAuth userAuth) {
...
+    // Get user email from UserService.
+    String email = userService.getCurrentUser().getEmail();
+
+    // Get user from datastore; if not present, add the user and return that.
+    Entity userEntity = getUserEntityFromEmail(email);
+    if (userEntity == null) {
+      Entity newUserEntity = addUserToDatabase(email);
+      return newUserEntity;
+    }
+
```
adamfweidman, line 143: nit: remove line

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,80 @@ private static String convertToJson(UserAuth userAuth) {
...
+    return userEntity;
+  }
+
+  // Return Entity object of user from database using email, or null if user 
+  // email is not in db.
```
adamfweidman, line 148: nit: format comment
/**
 *
 */

src/test/java/com/google/sps/AuthServletTest.java
```diff
@@ -200,6 +210,124 @@ public void testAddUserToDatabasePresent() throws Exception {
...
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> listResults = results.asList(FetchOptions.Builder.withDefaults());
+
+    // Verify that the Entity is in the database, and the Entity method return.
+    Assert.assertEquals(EMAIL, userEntityReturn.getProperty(User.USER_EMAIL));
+    Assert.assertEquals(1, listResults.size());
+
+    // Confirm that the Entity in the database matches the method return.
+    Assert.assertEquals(listResults.get(0), userEntityReturn);
   }
```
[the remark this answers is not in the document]
zpChris: If it's alright, I actually am going to keep the line if that's alright because I like a line separating the above line from the comment.

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: Yeah I did this for testing purposes, so I didn't have to use PowerMock to mock the result from the `UserServiceFactory.getUserService()`. However, now I think you're likely right -- I should just use PowerMock, so I'll change this.

adamfweidman: Whatever you think is best! I don't know the pros and cons of using PowerMock very well

zpChris: Changed!

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: Done!

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: Done!

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: I don't; I test it in a future CL, but I'll move those changes downstream to this branch.

zpChris: Actually, nevermind -- I do test this!

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,73 @@ private static String convertToJson(UserAuth userAuth) {
...
+      return newUserEntity;
+    }
+    return userEntity;
+  }
+
+  /**
+   * Return Entity object of user from database using email, or null if user 
+   * email is not in db.
+   */
+  public static Entity getUserEntityFromEmail(String email) {
```
tristan-jh, line 142: Will this method be used anywhere other than the tests? I'm wondering when a client would rather use this than the addUserToDatabase method.

src/test/java/com/google/sps/AuthServletTest.java
```diff
@@ -200,6 +219,136 @@ public void testAddUserToDatabasePresent() throws Exception {
...
+    Query query = new Query(User.USER);
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> listResults = results.asList(FetchOptions.Builder.withDefaults());
+
+    // Verify that the Entity is in the database, and the Entity method return.
+    Assert.assertEquals(EMAIL, userEntityReturn.getProperty(User.USER_EMAIL));
+    Assert.assertEquals(1, listResults.size());
+
+    // Confirm that the Entity in the database matches the method return.
+    Assert.assertEquals(listResults.get(0), userEntityReturn);
```
tristan-jh, line 264: is userEntityReturn also equal to userEntity created above? if so you can probably skip the part of this test which queries the db.

src/test/java/com/google/sps/AuthServletTest.java
```diff
@@ -200,6 +219,136 @@ public void testAddUserToDatabasePresent() throws Exception {
...
+    Entity userEntityReturn = AuthServlet.getUserEntityFromEmail(EMAIL);
+
+    // Retrieve the datastore results.
+    Query query = new Query(User.USER);
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> listResults = results.asList(FetchOptions.Builder.withDefaults());
+
+    // Verify that the Entity is in the database, and the Entity method return.
+    Assert.assertEquals(EMAIL, userEntityReturn.getProperty(User.USER_EMAIL));
+    Assert.assertEquals(1, listResults.size());
```
tristan-jh, line 261: optional: this block shows up often enough that it could be a helper method. eg: getSoleEntryInDatabase or similar. Not required though, redundancy in tests isn't too concerning as long as they're understandable (which these are).

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,73 @@ private static String convertToJson(UserAuth userAuth) {
...
+    // Get user email from UserService.
+    String email = userService.getCurrentUser().getEmail();
+
+    // Get user from datastore; if not present, add the user and return that.
+    Entity userEntity = getUserEntityFromEmail(email);
+    if (userEntity == null) {
+      Entity newUserEntity = addUserToDatabase(email);
+      return newUserEntity;
+    }
+    return userEntity;
```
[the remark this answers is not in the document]
zpChris: Ah you're right, good catch!

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,73 @@ private static String convertToJson(UserAuth userAuth) {
...
+      new FilterPredicate(User.USER_EMAIL, FilterOperator.EQUAL, email);
+    Query query = new Query(User.USER).setFilter(emailFilter);
     PreparedQuery results = datastore.prepare(query);
 
-    // If user is already in database, return.
-    for (Entity entity : results.asIterable()) {
-      if (entity.getProperty(User.USER_EMAIL).equals(email)) {
-        return;
-      }
+    // Get the list of results. If empty, return null; otherwise, return true.
```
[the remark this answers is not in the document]
zpChris: You're right, thank you! Done!

src/main/java/com/google/sps/servlets/AuthServlet.java
```diff
@@ -82,24 +88,73 @@ private static String convertToJson(UserAuth userAuth) {
   
   /**
    * If the user is not in database already, add them.
+   * This method returns the Entity object in the database (or the newly-created
+   * user Entity, if none previously existed).
    */
-  public void addUserToDatabase(String email) {
+  public static Entity addUserToDatabase(String email) {
```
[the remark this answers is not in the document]
zpChris: I think you're right, especially with the above change; done!

src/main/java/com/google/sps/servlets/AuthServlet.java
zpChris: The `addUserToDatabase` method (which is now named `getOrCreateUserInDatabase`) **calls** this method, `getUserEntityFromEmail`, so this is used there.

src/test/java/com/google/sps/AuthServletTest.java
zpChris: They are actually slightly different; after you add it to datastore, the Entity gets a **key** and **id**; the `userEntity` object does not have these attributes. I'm testing that these are added and that the Entity objects are equal with the final line.

src/test/java/com/google/sps/AuthServletTest.java
zpChris: If it's ok, I'll leave this block of code at least for now, as adding this block of code to a helper method would require a bit of syncing with the branches above (CLs stacked on top of this CL).

src/test/java/com/google/sps/AuthServletTest.java
tristan-jh: 👍

src/main/java/com/google/sps/servlets/AuthServlet.java
tristan-jh: ah sorry, I should have asked that differently. I'm wondering if this method can be private? or if you're expecting clients of this AuthServlet class to want to call getUserEntityFromEmail directly? If not then you could mark this protected (I think that'll still allow it to be called by tests) and/or add a comment indicating that this has non-private visibility only for use in tests.

zpChris: Oh I see -- yes, I think it's only used in tests! I'll confirm this and then change it, thanks for the note.

zpChris: Actually, I tried changing the access level to **protected**, but the tests can't access it then -- it throws an error when it tries to test it, and says `"getUserEntityFromEmail(java.lang.String) has protected access in com.google.sps.servlets.AuthServlet"`. This means I'll leave this as-is.

PR #33, opened 2020-07-16
Add Trip Entity objects to datastore upon homepage form submission.

zpChris:
### List of changes:
- Get form input from `/calculate-trip` POST request.
- Use the Google Maps Places Java Client API to get the `placeId` and **Place Details** of the `tripDestination`. This is used to get a **photo of the destination** and a **destination name**.
- Get the current logged-in User Entity from Datastore. Put the Trip Entity into Datastore, with the current logged-in User Entity as its ancestor.
- Add tests for the TripServlet class using Mockito and PowerMock (PowerMockito necessary for final and static methods).
- Redirect back to homepage if POST request is made with no user currently logged in.
- Add line in `.gitignore` to hide server-side `Config.java` file, which allows access to the Google Maps API in Java server-side files.

### Notes:
- In future CLs, lots of more code will be added to the `TripServlet.java` `doPost` method, so some merging decisions will need to be made that may lead to code restructuring.
- There is a possibility that the `tripKey` field of the Trip class should allowed to be `null`. This `tripKey` represents the **Key** of the Trip Entity in datastore, so if it is not yet in datastore, there should be no `tripKey`. However, I am operating with the assumption that the Trip should immediately be put in datastore before initialization, so I am leaving this unless there is an objection.
- I found that these methods may not be able to store POIs with non-English characters. This is interesting to note, but lower priority, so there's no action on this for now.
- The image source holds the API key (see the [Places API documentation](https://developers.google.com/places/web-service/photos#place_photo_requests) for why this is necessary).
- I resolved merge conflicts from master that get and post the events into / from Datastore in `TripServlet.java`; the trip form now redirects to the `"/trips/"` page after submission.

### Screencast:
[Storing Trip data in Datastore](https://screencast.googleplex.com/cast/NTIyMjE0MzU4OTYxMzU2OHw1NGFkNDAyYy01YQ)

### Images:
Image of local datastore service storing the Trip after input form submission (the API key in the photo URL is replaced with **HIDDEN_API_KEY** by directly editing the HTML of the page -- if you tested, you would see an actual API key there):
image
Image of the photo URL in the above datastore example:
[Image of the Island of Hawai'i](https://lh3.googleusercontent.com/p/AF1QipM7tbCZOj_5SOft9cYgI7un3bmieieqvdYkCPT5=s1600-w400)

### DIFF
--- a/src/main/java/com/google/sps/Trip.java
+++ b/src/main/java/com/google/sps/Trip.java
@@ -41,13 +41,13 @@ public class Trip {
   private int numDays;
 
   // Constants to get and put the Entity objects in Datastore.
-  private static final String TRIP = "trip";
-  private static final String TRIP_NAME = "trip_name";
-  private static final String DESTINATION_NAME = "destination_name";
-  private static final String TRIP_KEY = "trip_key";
-  private static final String IMAGE_SRC = "image_src";
-  private static final String START_DATE = "start_date";
-  private static final String END_DATE = "end_date";
+  public static final String TRIP = "trip";
+  public static final String TRIP_NAME = "trip_name";
+  public static final String DESTINATION_NAME = "destination_name";
+  public static final String TRIP_KEY = "trip_key";
+  public static final String IMAGE_SRC = "image_src";
+  public static final String START_DATE = "start_date";
+  public static final String END_DATE = "end_date";
 
   /**
    * Creates a new Trip.
--- a/src/main/java/com/google/sps/servlets/TripServlet.java
+++ b/src/main/java/com/google/sps/servlets/TripServlet.java
@@ -34,7 +34,8 @@ import javax.servlet.http.HttpServletResponse;
 public class TripServlet extends HttpServlet {
 
   // Create the GeoApiContext object.
-  GeoApiContext context;
+  private GeoApiContext context;
+  private int photoSrcSize = 400;
 
   // Constants to get form inputs.
   private static final String INPUT_TRIP_NAME = "inputTripName";
@@ -147,7 +148,7 @@ public class TripServlet extends HttpServlet {
         this.photoSrc = "../images/placeholder_image.png";
       } else {
         Photo photoObject = placeDetailsResult.photos[0];
-        this.photoSrc = getUrlFromPhotoReference(400, photoObject.photoReference);
+        this.photoSrc = getUrlFromPhotoReference(this.photoSrcSize, photoObject.photoReference);
       }
     }
   }
@@ -177,6 +178,10 @@ public class TripServlet extends HttpServlet {
    * Get a URL to show the photo from the photoreference.
    * See https://developers.google.com/places/web-service/photos#place_photo_requests
    * for more info.
+   * 
+   * @param maxWidth This is the maximum width of the image.
+   * @param photoReference This is the photo reference String stored in the 
+   * Google Maps Photo object; this is used to retrieve the actual photo URL.
    */
   public String getUrlFromPhotoReference(int maxWidth, String photoReference) {
     final String baseUrl = "https://maps.googleapis.com/maps/api/place/photo?";
--- a/src/main/java/com/google/sps/servlets/TripServlet.java
+++ b/src/main/java/com/google/sps/servlets/TripServlet.java
@@ -97,8 +97,6 @@ public class TripServlet extends HttpServlet {
     try {
       FindPlaceFromText findPlaceResult = findPlaceRequest.await();
 
-      System.err.println(findPlaceRequest);
-
       // Return place ID of the first candidate result.
       if (findPlaceResult.candidates != null) {
         return findPlaceResult.candidates[0].placeId;
--- a/src/main/java/com/google/sps/servlets/TripServlet.java
+++ b/src/main/java/com/google/sps/servlets/TripServlet.java
@@ -137,7 +137,7 @@ public class TripServlet extends HttpServlet {
     String destinationPlaceId = getPlaceIdFromTextSearch(context, this.tripDestination);
     if (destinationPlaceId == null) {
       this.destinationName = tripDestination;
-      this.photoSrc = "images/placeholder_image.png";
+      this.photoSrc = "../images/placeholder_image.png";
     } else {
       PlaceDetails placeDetailsResult = getPlaceDetailsFromPlaceId(context, destinationPlaceId);
 
--- a/src/main/java/com/google/sps/servlets/TripServlet.java
+++ b/src/main/java/com/google/sps/servlets/TripServlet.java
@@ -131,6 +131,11 @@ public class TripServlet extends HttpServlet {
   /**
    * Get the place ID of the text search. Return null if no place ID matches
    * the search.
+   * 
+   * @param context The entry point for making requests against the Google Geo 
+   * APIs (googlemaps.github.io/google-maps-services-java/v0.1.2/javadoc/com/google/maps/GeoApiContext.html).
+   * @param textSearch The text query to be entered in the findPlaceFromText(...)
+   * API call. Must be non-null.
    */ 
   public String getPlaceIdFromTextSearch(GeoApiContext context, String textSearch) 
     throws IOException {
@@ -261,6 +266,16 @@ public class TripServlet extends HttpServlet {
   /**
    * Store the Trip Entity in datastore with the User Entity as an ancestor.
    * Return the Trip Entity object.
+   * 
+   * @param response The HttpServletResponse used to redirect to homepage if
+   * no user is logged in. 
+   * @param tripName The human-readable name for the trip. Must be non-null.
+   * @param destinationName The name of the destination the user is heading to.
+   * This destination should be verified by the Google Maps API.
+   * @param tripDayOfTravel The date of the trip. Must be in yyyy-MM-dd date format.
+   * @param photoSrc The image source / URL to represent the trip. This is 
+   * typically retrieved using the Places API to get a photo from the destination
+   * name, but can also be the placeholder image source if no photo exists.
    */
   public Entity storeTripEntity(HttpServletResponse response, String tripName, 
     String destinationName, String tripDayOfTravel, String photoSrc) throws IOException {
--- a/src/main/java/com/google/sps/servlets/TripServlet.java
+++ b/src/main/java/com/google/sps/servlets/TripServlet.java
@@ -54,7 +54,7 @@ public class TripServlet extends HttpServlet {
 
   // Create the GeoApiContext object.
   private GeoApiContext context;
-  private int photoSrcSize = 400;
+  private static final int PHOTO_SRC_SIZE = 400;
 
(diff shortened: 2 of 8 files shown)


src/main/java/com/google/sps/Trip.java
```diff
@@ -41,13 +41,13 @@
...
-  private static final String IMAGE_SRC = "image_src";
-  private static final String START_DATE = "start_date";
-  private static final String END_DATE = "end_date";
+  public static final String TRIP = "trip";
+  public static final String TRIP_NAME = "trip_name";
+  public static final String DESTINATION_NAME = "destination_name";
+  public static final String TRIP_KEY = "trip_key";
+  public static final String IMAGE_SRC = "image_src";
+  public static final String START_DATE = "start_date";
+  public static final String END_DATE = "end_date";
```
eshika, line 50: Question: why were these changed to public? Is this for testing purposes?

zpChris: Yes, this is for testing purposes. In the next CL, I use the `Trip.TRIP` field variable to create a Query for the Trip Entity (`UserTripServlet.java` on CL #34), but yeah this is all just for use in `TripServletTest.java`.

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -18,16 +33,155 @@
...
+    try {
+      FindPlaceFromText findPlaceResult = findPlaceRequest.await();
+
+      // Return place ID of the first candidate result.
+      if (findPlaceResult.candidates != null) {
+        return findPlaceResult.candidates[0].placeId;
+      }
+      
+      // No candidate is given, so return null.
+      return null;
```
eshika, line 106: In what case would no candidate be given and no exception be thrown? Maybe we should throw an exception if no candidate is given so we don't run into a NullPointerException later on

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -18,16 +33,155 @@
...
+
+      // Get the name of the location from the place details result.
+      this.destinationName = placeDetailsResult.name;
+
+      // Get a photo of the location from the place details result.
+      if (placeDetailsResult.photos == null) {
+        this.photoSrc = "../images/placeholder_image.png";
+      } else {
+        Photo photoObject = placeDetailsResult.photos[0];
+        this.photoSrc = getUrlFromPhotoReference(400, photoObject.photoReference);
```
eshika, line 150: nit: save 400 as a constant variable (for easier change later on)

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -18,16 +33,155 @@
...
+    datastore.put(tripEntity);
+    return tripEntity;
+  }
+
+  /**
+   * Get a URL to show the photo from the photoreference.
+   * See https://developers.google.com/places/web-service/photos#place_photo_requests
+   * for more info.
+   */
+  public String getUrlFromPhotoReference(int maxWidth, String photoReference) {
```
eshika, line 181: Could you add a description of the parameters in the comment? I'm not sure what these params store

src/test/java/com/google/sps/TripServletTest.java
```diff
@@ -0,0 +1,275 @@
...
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query(User.USER);
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> listResults = results.asList(FetchOptions.Builder.withDefaults());
+
+    // Check whether the proper Entity and count were returned.
+    Assert.assertEquals(0, listResults.size());
+    Assert.assertNull(tripEntityReturn);
+  }
+
```
eshika, line 274: nit: remove extra line

src/main/java/com/google/sps/servlets/TripServlet.java
zpChris: I have a check for if the return is `null` -- check lines 135-138 on `TripServlet.java` (the same file). I'm thinking this could happen if, for some reason, the text search given by the user does not match any places that the backend can find.

While this is admittedly unlikely given I use the same API is an autocomplete, I figured it would probably be good to add a check; for instance, a malicious user could submit a text search that doesn't fit any actual location. I'm open to changing this into throwing an exception as you said though, because in that case we don't have a valid origin / destination.

Should I change this to throw some Exception if no candidate is given?

src/main/java/com/google/sps/servlets/TripServlet.java
zpChris: That's smart, I'll do that! Thanks 👍

src/main/java/com/google/sps/servlets/TripServlet.java
zpChris: Done!

src/test/java/com/google/sps/TripServletTest.java
zpChris: Done!

src/main/java/com/google/sps/servlets/TripServlet.java
eshika: Ok I understand! Returning null makes sense for the way you're using it because you can check if it's null and use a placeholder if so rather than stopping and throwing an exception, but my concern is if this method is used somewhere else (for example getting the place ID for routing), then it would cause a NullPointerException later on. Maybe we could have a default value instead?

zpChris: What sort of default value are you thinking? Also I think `null` might be better, as it lets each method that calls this method `getPlaceIdFromTextSearch(...)` to handle the result a unique way.

eshika: That reasoning makes sense, I'm fine with leaving this as is then!

zpChris:
I just resolved merge conflicts with master, and the changes are still the same (with a few very minor additions) and everything still compiles and works as expected.

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -41,6 +52,22 @@
 @WebServlet("/calculate-trip")
 public class TripServlet extends HttpServlet {
 
+  // Create the GeoApiContext object.
+  private GeoApiContext context;
+  private int photoSrcSize = 400;
```
tristan-jh, line 57: photoSrcSize could be a constant since it's never updated

src/test/java/com/google/sps/TripServletTest.java
```diff
@@ -0,0 +1,274 @@
...
+    // Retrieve the datastore results.
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query(Trip.TRIP);
+    PreparedQuery results = datastore.prepare(query);
+    List<Entity> listResults = results.asList(FetchOptions.Builder.withDefaults());
+
+    // Check that the Trip Entity has the User as an ancestor / parent.
+    Key userAncestorKey = listResults.get(0).getParent();
+    Entity userEntity = datastore.get(userAncestorKey);
```
tristan-jh, line 235: nice! this is really useful in a test b/c, in addition to testing the functionality, reading this clarifies how Trip and User entities are related to one another

src/test/java/com/google/sps/TripServletTest.java
```diff
@@ -0,0 +1,274 @@
...
+    // Mock UserService methods as logged-out user.
+    UserService userServiceMock = mock(UserService.class);
+    when(userServiceMock.isUserLoggedIn()).thenReturn(false);
+    when(userServiceMock.createLoginURL(AuthServlet.redirectUrl)).thenReturn(LOGIN_URL);
+
+    // PowerMock static getUserService() method, which is used to get the user.
+    PowerMockito.mockStatic(UserServiceFactory.class);
+    when(UserServiceFactory.getUserService()).thenReturn(userServiceMock);
+
+    // Run storeTripEntity(...), with the User logged in (so trip is stored).
```
tristan-jh, line 260: I'm a bit confused by this comment. I thought this test was to ensure that nothing is stored for not-logged-in users. And given that `isUserLoggedIn` is set to return false above my understanding was that the User is logged out during this call to `storeTripEntity`?

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -18,16 +33,160 @@
...
+  }
+
+  /**
+   * Store the Trip Entity in datastore with the User Entity as an ancestor.
+   * Return the Trip Entity object.
+   */
+  public Entity storeTripEntity(HttpServletResponse response, String tripName, 
+    String destinationName, String tripDayOfTravel, String photoSrc) throws IOException {
+    // Get User Entity. If user not logged in, redirect to homepage.
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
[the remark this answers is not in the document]
zpChris: Done!

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -18,16 +33,160 @@
...
+      return placeDetailsRequest.await();
+    } catch(ApiException | InterruptedException e) {
+      throw new IOException(e);
+    }
+  }
+
+  /**
+   * Populate the destinationName and photoSrc fields using the Google Maps API.
+   */
+  public void populateDestinationAndPhoto(GeoApiContext context, String tripDestination)
```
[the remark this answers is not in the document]
zpChris: Done!

src/main/java/com/google/sps/servlets/TripServlet.java
zpChris: Good point, thanks -- done!

src/test/java/com/google/sps/TripServletTest.java
zpChris: Thanks! :)

src/test/java/com/google/sps/TripServletTest.java
zpChris: You're right; I must've forgotten to update this comment, thanks for the note.

Fixed!

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -18,16 +33,160 @@
...
+      } else {
+        Photo photoObject = placeDetailsResult.photos[0];
+        this.photoSrc = getUrlFromPhotoReference(this.photoSrcSize, photoObject.photoReference);
+      }
+    }
+  }
+
+  /**
+   * Store the Trip Entity in datastore with the User Entity as an ancestor.
+   * Return the Trip Entity object.
```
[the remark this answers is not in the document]
zpChris: Ok that makes sense -- I added **@param** tags to **public** methods across the different Java files (except for `doGet` and `doPost`). Thanks for the note!

PR #34, opened 2020-07-17
Dynamically add trips for each user to the "/trips/" page.

zpChris:
### List of changes:
- Create `UserTripServlet.java` file with a `doGet(...)` function to retrieve the list of trips for the current user.
- Add a test to check the output of the `doGet(...)` function of the `UserTripServlet.java` file.
- Add multiple functions (`getAndAddTripCards(...)`, `addTripCards(...)`, and `buildTripCard(...)`) to the `tripScript.js` file to build and display the trip cards.
- The **Calendar** and **Maps** buttons on the trip cards now hold a query of the form `?tripKey=TRIP_KEY_HERE` to the links, which will allow the Calendar and Maps pages to show the relevant trips information. For now, however, this query is ignored.
- Remove static / constant Trip cards from the `/trips/` page.
- Revise styling of Bootstrap "cards" by placing `width: 15rem` in the `style.css` file, maintaining uniformity across the site.

### Notes:
- There are currently no restrictions on the length of the trip name, which means the "card" design format could be poor if a user gave a very long title. We may want to add restrictions on the trip name length in the future, but for now, it's not a big priority.
- The images on the trip cards are not always the same size, which can create a mismatch in card height. This is again a difficult issue to fix, and not high priority, so I am ignoring this for now.
- I don't know how the trip cards are ordered. However, it doesn't seem to be random, so I am not placing an emphasis on sorting these a certain way.

### Screencast:
[Dynamically Adding Trips for Each User](https://screencast.googleplex.com/cast/NTE2MzczODk4MDE1NTM5MnwwOWUyZWVhZS00MA)

### Images:
Image of `/trips/` page with no added trips:
image

Image of `/trips/` page with Hawaii trip added:
image

Image of `/trips/` page with four trips:
image

### DIFF
--- a/src/main/java/com/google/sps/servlets/UserTripServlet.java
+++ b/src/main/java/com/google/sps/servlets/UserTripServlet.java
@@ -77,5 +77,4 @@ public class UserTripServlet extends HttpServlet {
     String json = gson.toJson(tripList);
     return json;
   }
-
 }
--- a/src/main/java/com/google/sps/servlets/UserTripServlet.java
+++ b/src/main/java/com/google/sps/servlets/UserTripServlet.java
@@ -17,6 +17,7 @@ package com.google.sps.servlets;
 import com.google.appengine.api.datastore.DatastoreService;
 import com.google.appengine.api.datastore.DatastoreServiceFactory;
 import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.Key;
 import com.google.appengine.api.datastore.FetchOptions;
 import com.google.appengine.api.datastore.PreparedQuery;
 import com.google.appengine.api.datastore.Query;
@@ -29,8 +30,8 @@ import com.google.gson.Gson;
 import com.google.sps.data.User;
 import com.google.sps.Trip;
 import java.io.IOException;
-import java.util.List;
 import java.util.ArrayList;
+import java.util.List;
 import javax.servlet.annotation.WebServlet;
 import javax.servlet.http.HttpServlet;
 import javax.servlet.http.HttpServletRequest;
@@ -53,9 +54,24 @@ public class UserTripServlet extends HttpServlet {
       return;
     }
 
+    // Write the list of Trips to file through the HttpServletResponse object.
+    writeTripsToFile(response, userEntity.getKey());
+  }
+
+  /**
+   * Retrieve the list of Trips from datastore using the User Entity Key, and 
+   * write the Trips to file using the HttpServletResponse object.
+   *
+   * @param response The HttpServletResponse object which is used to write the
+   * Trips to the file. Must be non-null.
+   * @param userEntityKey The Key of the User Entity which is currently signed 
+   * in; this User Entity should be the ancestor / parent of the Trip Entity
+   * objects. Must be non-null.
+   */
+  public void writeTripsToFile(HttpServletResponse response, Key userEntityKey) throws IOException {
     // Get the list of Trip Entity objects through a Query.
     DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
-    Query query = new Query(Trip.TRIP, userEntity.getKey());
+    Query query = new Query(Trip.TRIP, userEntityKey);
     PreparedQuery results = datastore.prepare(query);
 
     // Iterate over the trip Entity objects, and convert them to Trip objects.
--- a/src/main/java/com/google/sps/servlets/UserTripServlet.java
+++ b/src/main/java/com/google/sps/servlets/UserTripServlet.java
@@ -1,0 +1,81 @@
+// Copyright 2019 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package com.google.sps.servlets;
+
+import com.google.appengine.api.datastore.DatastoreService;
+import com.google.appengine.api.datastore.DatastoreServiceFactory;
+import com.google.appengine.api.datastore.Entity;
+import com.google.appengine.api.datastore.FetchOptions;
+import com.google.appengine.api.datastore.PreparedQuery;
+import com.google.appengine.api.datastore.Query;
+import com.google.appengine.api.datastore.Query.Filter;
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.gson.Gson;
+import com.google.sps.data.User;
+import com.google.sps.Trip;
+import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/user-trips")
+public class UserTripServlet extends HttpServlet {
+
+  // This allows the doGet method to return null.
+  private static final String nullReturn = null;
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("application/json;");
+
+    // If no user is signed in, return null.
+    Entity userEntity = AuthServlet.getCurrentUserEntity(AuthServlet.getUserService());
+    if (userEntity == null) {
+      response.getWriter().println(nullReturn);
+      return;
+    }
+
+    // Get the list of Trip Entity objects through a Query.
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
+    Query query = new Query(Trip.TRIP, userEntity.getKey());
+    PreparedQuery results = datastore.prepare(query);
+
+    // Iterate over the trip Entity objects, and convert them to Trip objects.
+    List<Trip> tripList = new ArrayList<>();
+    for (Entity entity : results.asIterable()) {
+      tripList.add(Trip.buildTripFromEntity(entity));
+    }
+
+    // Convert the trip list to JSON, and return JSON. Empty list signals no trips.
+    String json = convertTripListToJson(tripList);
+    response.getWriter().println(json);
+  }
+
+  /**
+  * Converts a List of Trips into a JSON string using the Gson library.
+  */
+  private String convertTripListToJson(List<Trip> tripList) {
+    Gson gson = new Gson();
+    String json = gson.toJson(tripList);
+    return json;
+  }
+
+}
--- a/src/main/java/com/google/sps/servlets/UserTripServlet.java
+++ b/src/main/java/com/google/sps/servlets/UserTripServlet.java
@@ -47,7 +47,7 @@ public class UserTripServlet extends HttpServlet {
     response.setContentType("application/json;");
 
     // If no user is signed in, return null.
-    Entity userEntity = AuthServlet.getCurrentUserEntity(AuthServlet.getUserService());
+    Entity userEntity = AuthServlet.getCurrentUserEntity();
     if (userEntity == null) {
       response.getWriter().println(nullReturn);
       return;
--- a/src/main/webapp/trips/tripScript.js
+++ b/src/main/webapp/trips/tripScript.js
@@ -21,9 +21,86 @@
 $(document).ready(() => {
   // Set the content width of the site.
   setContentWidth('800px');
+
+  // Add the trip cards to the site.
+  getAndAddTripCards();
 });
 
 // Set the width of the content container.
 function setContentWidth(width) {
   document.getElementById('content').style.width = width;
+}
+
+// Fetch the trip details for the user from the database, add trip cards.
+function getAndAddTripCards() {
+  fetch('/user-trips').then(response => response.json()).then((trips) => {
+    // If trips are present, add them to the page.
+    if (trips !== null) {
+      addTripCards(trips);
+    }
+  }).catch((error) => {
+    console.error(error);
+  });
+}
+
+// Add the trip cards to the site.
+function addTripCards(trips) {
+  trips.forEach((trip) => {
+    const tripCardContainer = document.getElementById('trips-cards-container');
+
+    // Build and add the trip card.
+    const tripCard = buildTripCard(trip.tripName, trip.destinationName, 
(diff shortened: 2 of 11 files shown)


src/main/java/com/google/sps/servlets/UserTripServlet.java
```diff
@@ -0,0 +1,81 @@
...
+  private static final String nullReturn = null;
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    response.setContentType("application/json;");
+
+    // If no user is signed in, return null.
+    Entity userEntity = AuthServlet.getCurrentUserEntity();
+    if (userEntity == null) {
+      response.getWriter().println(nullReturn);
```
adamfweidman, line 52: Do you even need to write anything out? Could you just return? Or is null used somewhere in the frontend

src/main/webapp/trips/tripScript.js
```diff
@@ -21,9 +21,86 @@
...
+  getAndAddTripCards();
 });
 
 // Set the width of the content container.
 function setContentWidth(width) {
   document.getElementById('content').style.width = width;
 }
+
+// Fetch the trip details for the user from the database, add trip cards.
+function getAndAddTripCards() {
```
adamfweidman, line 35: I take my previous comment back I see that null is used here!

src/test/java/com/google/sps/TripTest.java
```diff
@@ -69,6 +97,36 @@ public void testTripConstructorSingleDay() {
...
+    datastore.put(tripEntity);
+    Trip trip = Trip.buildTripFromEntity(tripEntity);
+
+    Assert.assertEquals(trip.getTripName(), TRIP_NAME);
+    Assert.assertEquals(trip.getDestinationName(), DESTINATION_NAME);
+    Assert.assertEquals(trip.getImageSrc(), IMAGE_SRC);
+    Assert.assertEquals(trip.getStartDate(), START_DATE);
+    Assert.assertEquals(trip.getEndDate(), END_DATE);
+  }
+
```
adamfweidman, line 129: This is probably how I should have tested my event stuff as well but I actually just converted in from object --> entity --> object and then confirmed that those were the same. No change required just kind of thinking out loud.

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(UserServiceFactory.class)
+public final class UserTripServletTest {
+
+  // Add constants necessary to testing retrieval of the current User.
+  public static final String EMAIL = "testemail@gmail.com";
+  public static final String AUTH_DOMAIN = "gmail.com";
+  public static final String LOGOUT_URL = "/_ah/logout?continue=%2F";
+  public static final String LOGIN_URL = "/_ah/login?continue=%2F";
```
adamfweidman, line 50: unless you use these outside they should be either private or protected!

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -0,0 +1,166 @@
...
+
+  // Create UserTripServlet and TripServlet objects.
+  UserTripServlet userTripServlet;
+  TripServlet tripServlet;
```
adamfweidman, line 54: make these private

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -0,0 +1,166 @@
...
+    // Create Trip Entity properties.
+    final String tripName = "Family Vacation";
+    final String destinationName = "Island of Hawai'i";
+    final String tripDayOfTravel = "2020-07-17";
+    final String photoSrc = "../images/placeholder_image.png";
+
+    // Mock UserService methods as logged-in user.
+    UserService userServiceMock = mock(UserService.class);
+    when(userServiceMock.isUserLoggedIn()).thenReturn(true);
+    // This is the User object from Google Appengine (full path given to avoid
```
adamfweidman, line 92: Jut to be consistent with the rest of your code add a line between lines 91 and 92

src/main/webapp/trips/tripScript.js
```diff
@@ -21,9 +21,86 @@
...
+  titleElement.innerText = tripTitle;
+
+  const datesDestinationElement = document.createElement('p');
+  datesDestinationElement.className = 'card-text';
+  datesDestinationElement.innerText = destinationName + ', ' + startDate.year + 
+    '-' + startDate.month + '-' + startDate.day;
+
+  const calendarButton = document.createElement('a');
+  calendarButton.className = 'btn btn-primary trip-button';
+  calendarButton.href = '../calendar.html?tripKey=' + tripKey;
```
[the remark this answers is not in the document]
zpChris: Yes it is! It is a long CL, sorry about that, but you're exactly right.

src/main/java/com/google/sps/servlets/UserTripServlet.java
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+  * Converts a List of Trips into a JSON string using the Gson library.
+  */
+  private String convertTripListToJson(List<Trip> tripList) {
+    Gson gson = new Gson();
+    String json = gson.toJson(tripList);
+    return json;
+  }
+
```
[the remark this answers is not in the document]
zpChris: Done!

src/main/java/com/google/sps/servlets/UserTripServlet.java
zpChris: That's a good question! I'm slightly concerned because if there are no trips, then you could reasonably return nothing, but that doesn't mean the user isn't logged in; that just means there are no trips saved. If no user exists, there is no way to even search for trips, so I want to differentiate those return values clearly.

src/main/webapp/trips/tripScript.js
zpChris: Yes great, that's right! :)

src/test/java/com/google/sps/TripTest.java
zpChris: Ok, gotcha 👍

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: Done!

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: You're right, thanks -- fixed!

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: Fixed!

src/main/java/com/google/sps/servlets/UserTripServlet.java
```diff
@@ -0,0 +1,80 @@
...
+    // If no user is signed in, return null.
+    Entity userEntity = AuthServlet.getCurrentUserEntity();
+    if (userEntity == null) {
+      response.getWriter().println(nullReturn);
+      return;
+    }
+
+    // Get the list of Trip Entity objects through a Query.
+    DatastoreService datastore = DatastoreServiceFactory.getDatastoreService();
```
bheff-work, line 57: Has Adam talked to you about creating utility methods for these DatastoreService calls? These seem to be repeating a lot throughout the codebase, and I think it might be a good idea to create methods that make these calls so you, Eshika, and Adam don't all have to write the same calls with slight differences.

Not only will it make it easier to read/write, but will also make the code stronger, as whenever you want to change how these calls are made or how the elements are stored, you'll only have to make changes in one place instead of 3.

I'm not necessarily going to make you write these changes now, but I really want you three to chat about this and come up with a plan forward

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -0,0 +1,167 @@
...
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import org.junit.After;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import static org.mockito.Mockito.*;
+import org.powermock.api.mockito.PowerMockito;
```
bheff-work, line 38: static imports should go above non static ones

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -0,0 +1,167 @@
...
+    helper.setUp();
+  }
+
+  @After
+  public void tearDown() {
+    helper.tearDown();
+  }
+
+  @Test
+  public void testDoGetResponseLoggedIn() throws Exception {
```
bheff-work, line 78: I feel like there are some test cases you haven't covered here e.g. no trips returned from datastore and multiple trips returned from datastore

bheff-work, line 78: These tests are getting pretty big, and though it's often good to use real objects, but there is definitely a line.  Your test is relying on the real behavior of both the TripServlet and the DatastoreService which means you're adding a *lot* of logic to your tests and putting yourself in a position of having your unit tests fail because of things entirely out of your control. 

I think you need to restrict the scope of these tests a bit to focus more on UserTripServlet

src/main/java/com/google/sps/servlets/UserTripServlet.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.appengine.api.datastore.Query.FilterOperator;
+import com.google.appengine.api.datastore.Query.FilterPredicate;
+import com.google.appengine.api.users.UserService;
+import com.google.appengine.api.users.UserServiceFactory;
+import com.google.gson.Gson;
+import com.google.sps.data.User;
+import com.google.sps.Trip;
+import java.io.IOException;
+import java.util.List;
+import java.util.ArrayList;
```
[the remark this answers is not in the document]
zpChris: Fixed! Thanks, I'll remember to keep `ArrayList` before `List` in the imports.

src/main/java/com/google/sps/servlets/UserTripServlet.java
zpChris: Yes, Adam mentioned it! We (Eshika, Adam, and I) still have to hop on a call to talk about this, and can hopefully do so late today or tomorrow morning.

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: I didn't know that, thanks for the note -- done!

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: That's true, I'll be sure to add these tests.

zpChris: Ok, that makes sense -- I'll restructure the `UserTripServlet.java` file so that you can test the return if you already have the User Entity.

Should I have one test for the full `doGet(...)` method (like an integration test), or is that unnecessary?

zpChris: I restructured the `UserTripServlet.java` file and tested a new method, `writeTripsToFile(...)`, which takes in the `HttpServletResponse` object and the `Key userEntityKey`, and writes the Trips in JSON through the `response` object. I also included helper methods in `UserTripServletTest.java` that set up the Datastore with zero, one, or two trips.

zpChris: I added these, and removed the tests that check whether or not the user is logged in, as these aren't necessary anymore as a result of my new `UserTripServlet.java` file structure.

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -49,9 +53,21 @@
   private static final String LOGOUT_URL = "/_ah/logout?continue=%2F";
   private static final String LOGIN_URL = "/_ah/login?continue=%2F";
 
-  // Create UserTripServlet and TripServlet objects.
+  // Cosntants to represent different Trip attributes.
```
bheff-work, line 56: nit: "Constants"

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -74,94 +89,167 @@ public void tearDown() {
     helper.tearDown();
   }
 
+  /**
+   * Add a User Entity and no Trip Entity objects to Datastore, and return the
```
bheff-work, line 93: what is a "no Trip Entity"?

src/test/java/com/google/sps/UserTripServletTest.java
```diff
@@ -74,94 +89,167 @@ public void tearDown() {
...
 
     // Create writers to pass into the mock response object.
     StringWriter stringWriter = new StringWriter();
     PrintWriter writer = new PrintWriter(stringWriter);
     when(responseMock.getWriter()).thenReturn(writer);
 
-    // Run the UserServlet doGet(...) method with request and response mocks. 
-    userTripServlet.doGet(requestMock, responseMock); 
+    // Get the User Entity Key, and set up the database using the helper method.
+    Key userEntityKey = addCurrentUserAndNoTripDatastore();
```
bheff-work, line 179: Though it may seem counter-intuitive, in tests it's generally discouraged to use helper methods unless something is being done a bunch of times (and in some languages at Google, such as c++, even that is discouraged). The reasoning is that with tests it's important that you understand exactly what's happening in the test, to ensure that nothing is being obfuscated. When people read helper methods in tests, it's easy to just glide your eyes over them instead of hunting down what's happening in that method.

So though it makes the test itself longer, you probably should stick these helper methods back into the tests they come from.

I know it goes against what we do in regular code, so let me know if you're still confused by this

bheff-work: I think I would be cool with you making a populateTripEntity helper method, since that is pretty simple setting code, and will be re-used multiple places, and is easy to digest

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: Oh that just means that I'm not adding Trip Entity objects to Datastore. I'll clarify this in the comment.

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: The `populateTripEntity` helper method would be the same `buildEntity` method in the `Trip.java` class. Can I just use that @bheff-work?

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: This is not applicable anymore, as I deleted this helper method.

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: Actually, I didn't want to over-complicate things, so I just included all of this code in each test. There are only three tests here, so it's not repeated too many times.

src/test/java/com/google/sps/UserTripServletTest.java
zpChris: Thanks for the note!

PR #36, opened 2020-07-29
Create a pop-up when an event is clicked on the calendar page

adamfweidman:
### **Major changes:**

- In order to create the map I needed to add a `placeId` field to the **Event.java** class (which I will need to use later on anyway)

- **calendar-script.js**: 
-- add the `eventClick()` function which calls `createMap()` which dynamically creates everything that goes in popup (called modal in code)
-- In order to display the open and closing hours for the specific date I needed to convert from hhmm (24hr) format to AM/PM format -- EG: 2320 --> 11:20 pm (this happens in the `formatAMPM()` function). If no open hours are available "open all day" is displayed

- **TripServletTest.java** use powermock to mock the _placeIdFromAddress_ functionality

**Example of Specific Opening hours:**
Screenshot 2020-07-28 at 6 40 26 PM

**Example of "All Day" opening hours:**
Screenshot 2020-07-28 at 6 53 47 PM


pom.xml
```diff
@@ -93,31 +93,6 @@
...
-      <artifactId>powermock-module-junit4</artifactId>
-      <version>2.0.0-beta.5</version>
-    </dependency>
-
-    <dependency>
-      <groupId>com.google.maps</groupId>
-      <artifactId>google-maps-services</artifactId>
-      <version>0.14.0</version>
-    </dependency>
-
```
zpChris, line 120: Thanks for removing these!

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -227,18 +227,17 @@ public Entity putTripDayInDatastore(String origin, DatastoreService datastore, L
     for (String address : pois) {
       // create event entity
       String name = address.split(",")[0];
-      Event event = new Event(name, address, startDateTime, travelTimes.get(travelTimeIndex));
+      String placeId = getPlaceIdFromTextSearch(this.context, address);
+      Event event = new Event(name, address, placeId, startDateTime, HALF_HOUR);
```
zpChris, line 231: Why is this just `HALF_HOUR` now, and not `travelTimes.get(travelTimeIndex)`?

src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -227,18 +227,17 @@ public Entity putTripDayInDatastore(String origin, DatastoreService datastore, L
...
       Entity eventEntity = event.eventToEntity(tripDayEntity.getKey());
       eventEntities.add(eventEntity);
 
-      // put entity in datastore   
+      // put entity in datastore
```
zpChris, line 235: nit: Remove spaces? Looks like nothing was changed here.

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
     });
   });
 }
+
+function createMap(modalBody, eventObj) {
+  const infoDisplay = document.createElement('div');
+  let address = document.createElement('p');
+  var urlFormatAddress = eventObj.extendedProps.address.replace(/\s/g, '+')
+  var url = 'https://www.google.com/maps/search/?api=1&query=' + urlFormatAddress + '&query_place_id=' + eventObj.extendedProps.placeId;
+  infoDisplay.innerHTML = '<b>Address: </b><a href=' + url + '>' + eventObj.extendedProps.address + '</a>' + '<br>';
```
zpChris, line 87: nit / personal preference: Add `target="_blank"` as an attribute to the `a` tag so that it opens in a new tab (https://www.w3schools.com/tags/att_a_target.asp). This way you still have the website open.

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+  infoDisplay.appendChild(address);
+  modalBody.appendChild(infoDisplay)
```
zpChris, line 89: I don't think this code is doing what you expect. The `address` `p` element / object is never actually used, but it's added to `infoDisplay` -- all this is doing is adding a space between the address line and the open / close lines.

Also, you are setting `infoDisplay.innerHTML` in order to add the bolded "Address:" and the `a` tag. My preference would be to actually build the elements in JavaScript, and append them as children like you do in the rest of this function. I'll let you make that decision though, since I'm not sure if one is particularly more proper than the other.

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+          openTime: '9AM',
+          closeTime: '5PM',
+        }
       });
     });
   });
 }
+
+function createMap(modalBody, eventObj) {
+  const infoDisplay = document.createElement('div');
```
zpChris, line 83: This `div` (infoDisplay) isn't actually doing anything. Unless you add styling to this, you should take it out and add the `innerHTML` of `infoDisplay` directly to the `modalBody`.

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+  modalBody.appendChild(infoDisplay)
+
+  // create new div to hold map
+  const mapDis = document.createElement('div');
+  mapDis.id = 'map';
+  modalBody.appendChild(mapDis);
+
+  // instantiate map
+  map = new google.maps.Map(document.getElementById('map'), {
+      zoom: 13,
```
zpChris, line 98: Use a constant for this zoom factor of **13**.

src/main/webapp/style.css
```diff
@@ -141,3 +141,8 @@
 .trip-button {
   margin-right: 5px;
 }
+
+#map {
+  height: 200px;
```
zpChris, line 146: Change the height to `250px`. At `200px`, the "fullscreen" button on the Maps is only partially showing.

`200px` (see the two lines above the little person):
image

`250px` (see the fullscreen button above the little person):
image

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+      center: new google.maps.LatLng(0, 0)
+    });
```
zpChris, line 100: nit: Fix spacing (the zoom, center, and end lines have one extra tab).

src/main/webapp/calendar-script.js
```diff
@@ -12,8 +12,17 @@
...
 // limitations under the License.
 
+let script = document.createElement('script');
+script.src = 'https://maps.googleapis.com/maps/api/js?key=' + config.API_KEY + 
+  '&libraries=places&callback=initMod';
+script.defer = true;
+script.async = true;
+
+var map;
+var service;
```
zpChris, line 22: Add a comment to describe what these are for.

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+
+  // create new div to hold map
+  const mapDis = document.createElement('div');
+  mapDis.id = 'map';
+  modalBody.appendChild(mapDis);
+
+  // instantiate map
+  map = new google.maps.Map(document.getElementById('map'), {
+      zoom: 13,
+      center: new google.maps.LatLng(0, 0)
```
zpChris, line 99: Do you have to set a value for the `center` attribute initially? I removed it on the local test and nothing changed, so I think you can remove it.

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+    });
+
+  service = new google.maps.places.PlacesService(map);
+  service.getDetails({
+    placeId: eventObj.extendedProps.placeId
+  }, function(result, status) {
+    if (status != google.maps.places.PlacesServiceStatus.OK) {
+      alert(status);
+      return;
+    }
```
zpChris, line 109: No action required, just a question: Do you know when this would fail? When would this if statement be true?

src/main/webapp/calendar-script.js
zpChris: It seems like these are just used within the `createMap(...)` function -- can you move them there and change `var` to `const`?

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+
+    // if there are open hours display them, otherwise open all day
+    try {
+      var stringFullDate = eventObj.extendedProps.stringDate;
+      var dateStr = stringFullDate.split('T')[0]; 
+      var intOfWeek = getintOfWeek(dateStr);
+
+      var dateObj = new Date(stringFullDate);
+      var openTime = formatAMPM(result.opening_hours.periods[intOfWeek].open.time);
+      var closeTime = formatAMPM(result.opening_hours.periods[intOfWeek].close.time);
```
zpChris, line 124: I know this is late as we already have code in the database that doesn't follow this rule, but the JavaScript style guide notes to use `let` or `const` instead of `var`: https://google.github.io/styleguide/jsguide.html#features-use-const-and-let. This is possibly a nit, I'll let you decide what to do here.

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,95 @@ function getEvents(calendar) {
...
+
+    var marker = new google.maps.Marker({
+      map: map,
+      position: result.geometry.location
+    });
+  });
+}
+
+// takes a string in date format and return the int of the week
+function getintOfWeek(date) {
```
zpChris, line 141: nit: CamelCase. i.e.; `getIntOfWeek`.

src/main/webapp/calendar-script.js
zpChris: There are other `var`s in your codebase, but I won't leave this comment on the rest of them.

src/main/webapp/calendar.html
```diff
@@ -3,14 +3,18 @@
   <head>
     <meta charset='utf-8' />
     <title>TravIS Calendar</title>
-    <link href='lib/fullcalendar/main.css' rel='stylesheet' />
+    <link href='lib/fullcalendar/main.css' rel='stylesheet'/>
```
zpChris, line 6: nit: There's no change here, and I think convention is to have a space, so I know this is really nitpicky but my nit is to change that back.

src/test/java/com/google/sps/CalendarServletTest.java
```diff
@@ -173,7 +175,7 @@ public void testWriteEventsMulTripDays() throws Exception {
             "\"travelTime\":30}]";
     
     writer.flush();
-    Assert.assertTrue(stringWriter.toString().contains(expectedJson));
+    //Assert.assertTrue(stringWriter.toString().contains(expectedJson));
```
zpChris, line 178: Uncomment this out.

src/test/java/com/google/sps/EventTest.java
```diff
@@ -23,6 +23,7 @@
 import com.google.appengine.tools.development.testing.LocalDatastoreServiceTestConfig;
 import com.google.appengine.tools.development.testing.LocalServiceTestHelper;
 import com.google.sps.data.Event;
+import java.io.IOException;
```
zpChris, line 26: Why did this need to be added?

src/test/java/com/google/sps/TripServletTest.java
```diff
@@ -243,6 +246,46 @@ public void testPutTripDayInDatastore() throws Exception {
...
+
+    // manually create params list
+    List<String> paramsList = new ArrayList<>();
+    paramsList.add("inputDestination");
+    paramsList.add("inputDayOfTravel");
+    paramsList.add("inputTripName");
+    paramsList.add("poi-1");
+    paramsList.add("poi-2");
+    Enumeration<String> params = Collections.enumeration(paramsList);
+
```
zpChris, line 288: Note: It'll likely be a good idea to put some of this into the Utility class.

adamfweidman: Agreed!

src/main/webapp/calendar-script.js
adamfweidman: i'm not sure but its in all the online examples that checks if theres an error getting the information

src/main/webapp/calendar-script.js
adamfweidman: removed them all

src/main/webapp/calendar-script.js
adamfweidman: I need a div because the map is being added before the open hours can be

adamfweidman: i need it because i am adding the hours to the modalBody after i add the address and map. just the way it is structured. so if i want the open hours to be above the map I need the div

zpChris: OK gotcha, no problem 👍

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,125 @@ function getEvents(calendar) {
...
+}
+
+// converts from 24hr (hh:mm) to 12hr format (hh:mm AM/PM)
+function formatAMPM(time) {
+  let hours = time.substring(0, 2);
+  let minutes = time.substring(2, 4);
+  const ampm = hours >= 12 ? 'pm' : 'am';
+  hours = hours % 12;
+  hours = hours ? hours : 12; // the hour '0' should be '12'
+  minutes = (minutes < 10 && minutes > 0) ? '0' + minutes : minutes;
```
bheff-work, line 184: why is this conversion necessary? if you're going from 24hr hh:mm to 12hr hh:mm,  the substring you get on line 180 should already have the '0' in front, no?

adamfweidman: Yes you are right! I was thinking in Java where parsing "03" from a string converts it to 3

src/main/webapp/calendar-script.js
```diff
@@ -43,8 +66,125 @@ function getEvents(calendar) {
...
+      return;
+    }
+    // set center of map
+    map.setCenter(result.geometry.location);
+  
+    // add open hours to mod info Display
+    let openHours = document.createElement('p');
+
+    // if there are open hours display them, otherwise open all day
+    try {
```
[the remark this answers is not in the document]
adamfweidman: only `openTime` and `closeTime` will throw an exception but if I create the HTML objects outside the try then its possible that it may display ```Open All Day``` ```Open: undefined, Close:undefinied```

adamfweidman:
Additional functionality added to pop-up after CL:
The address became a link to a google maps page centered at the location of the event (in a new tab).

PR #38, opened 2020-07-29
Add the index.yaml file.

zpChris:
This CL adds the `index.yaml` file which sets the configuration for the **Location** Entity.

### DIFF
--- a/src/main/webapp/WEB-INF/index.yaml
+++ b/src/main/webapp/WEB-INF/index.yaml
@@ -1,0 +1,7 @@
+indexes:
+
+- kind: location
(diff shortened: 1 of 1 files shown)


src/main/webapp/WEB-INF/index.yaml
```diff
@@ -0,0 +1,7 @@
+indexes:
+
```
tristan-jh, line 2: optional: add a short comment explaining where this is used. Eg: "This index need when querying ... etc".

zpChris: Done!

PR #39, opened 2020-07-31
Get and store end date input for multi day trips.

eshika:
This CL enables the user to enter an end date and stores that end date as part of the Trip entity in Datastore.

### DIFF
--- a/src/main/java/com/google/sps/servlets/TripServlet.java
+++ b/src/main/java/com/google/sps/servlets/TripServlet.java
@@ -81,6 +81,7 @@ public class TripServlet extends HttpServlet {
   private String tripName;
   private String tripDestination;
   private String tripDayOfTravel;
+  private String tripEndDate;
   private String destinationName;
   private String photoSrc;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/google/sps/servlets/TripServlet.java
```diff
@@ -81,6 +81,7 @@
   private String tripName;
   private String tripDestination;
   private String tripDayOfTravel;
```
zpChris, line 83: Change this variable to `tripStartDate` to match the new format of start / end date instead of `tripDayOfTravel`.

zpChris: There are other instances of this -- I won't leave a comment for every instance though.

eshika: Decided against this as this variable appears in multiple places and in multiple forms (e.g. INPUT_DAY_OF_TRAVEL and tripDayOfTravel)

PR #40, opened 2020-07-31
Add popups and markers.

eshika:
This CL makes the map on the maps page more intuitive for the user by distinguishing the hotel marker from the POI markers and adding info popups to each of the markers.

Note: Because I am now querying the locations and adding the markers based on the place IDs from that query, not all the markers are perfectly on the route. However, the marker locations are still in the vicinity and tend to correspond to the central point of the location so I believe this is fine.

### DIFF
--- a/src/main/webapp/mapsScript.js
+++ b/src/main/webapp/mapsScript.js
@@ -157,7 +157,7 @@ function addInfoWindow(results, marker) {
   const urlFormatAddress = results[0].name.replace(/\s/g, '+');
   const url = 'https://www.google.com/maps/search/?api=1&query=' + urlFormatAddress 
               + '&query_place_id=' + results[0].placeId;
-  contentString += "<a href=\""+url+"\">View on Google Maps</a>";
+  contentString += "<a href=\""+url+"\" target=\"_blank\">View on Google Maps</a>";
   // Construct infoWindow.
   const infoWindow = new google.maps.InfoWindow({
     content: contentString,
--- a/src/main/webapp/mapsScript.js
+++ b/src/main/webapp/mapsScript.js
@@ -23,9 +23,8 @@ var map;
 var directionsService;
 var directionsRenderer;
 
-// Map marker label constants
+// Map marker label constant
 var labels = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
-var labelIndex = 1;
 
 // Attach callback function to the `window` object
 window.initMap = function() {
--- a/src/main/webapp/mapsScript.js
+++ b/src/main/webapp/mapsScript.js
@@ -151,16 +151,32 @@ function addPoiMarker(location, isOrigin, index) {
  */
 function addInfoWindow(results, marker) {
   // Format and add place name, address, and link to Google Maps.
-  let contentString = "";
-  contentString += "<b>" + results[0].name + "</b>";
-  contentString += "<p>" + results[0].formatted_address + "</p>";
+  const info = document.createElement('div');
+  info.id = 'infoWindow';
+
+  // Add place name in bold.
+  const boldName = document.createElement('b');
+  boldName.innerText = results[0].name;
+  info.appendChild(boldName);
+
+  // Add formatted address.
+  const address = document.createElement('p');
+  address.innerText = results[0].formatted_address;
+  info.appendChild(address);
+
+  // Add link to Google Maps.
   const urlFormatAddress = results[0].name.replace(/\s/g, '+');
   const url = 'https://www.google.com/maps/search/?api=1&query=' + urlFormatAddress 
               + '&query_place_id=' + results[0].placeId;
-  contentString += "<a href=\""+url+"\" target=\"_blank\">View on Google Maps</a>";
+  const link = document.createElement('a');
+  link.href = url;
+  link.target = '_blank';
+  link.innerText = 'View on Google Maps';
+  info.appendChild(link);
+
   // Construct infoWindow.
   const infoWindow = new google.maps.InfoWindow({
-    content: contentString,
+    content: info,
     maxWidth: 200
   });
   // Open/close infoWindow on marker click.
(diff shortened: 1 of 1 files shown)


eshika:
Here's a screenshot of what the new markers look like (hotel/origin in green).

image

eshika:
Screenshow of popup windows: 

image

eshika:
Clicking on the Google Maps link leads to opening a page like this in a new tab:

image

src/main/webapp/mapsScript.js
```diff
@@ -90,6 +93,93 @@ function showDirections(locations) {
       window.alert('Directions request failed due to ' + status);
     }
   });
+
+  // Add markers to map.
+  addPoiMarker(origin, true, 0);  
+  for (let i = 0; i < waypts.length; i++) {
+    let index = i + 1;
+    addPoiMarker(waypts[i].location, false, index);
+  }
```
eshika, line 102: Note: I know that I am looping through the waypts twice (once to parse and once to add markers), however I have to add these markers after the directions have been calculated so I can query for their place IDs in the vicinity of the map center (once the directionsRenderer has recentered the map).

src/main/webapp/mapsScript.js
```diff
@@ -90,6 +93,93 @@ function showDirections(locations) {
...
+ */
+function addInfoWindow(results, marker) {
+  // Format and add place name, address, and link to Google Maps.
+  let contentString = "";
+  contentString += "<b>" + results[0].name + "</b>";
+  contentString += "<p>" + results[0].formatted_address + "</p>";
+  const urlFormatAddress = results[0].name.replace(/\s/g, '+');
+  const url = 'https://www.google.com/maps/search/?api=1&query=' + urlFormatAddress 
+              + '&query_place_id=' + results[0].placeId;
+  contentString += "<a href=\""+url+"\" target=\"_blank\">View on Google Maps</a>";
```
adamfweidman, line 160: If you're creating an HTML object, you should probably do it dynamically in Javascript. `document.createElement('p')` etc... You can look at my PopUp branch, I changed mine as well

src/main/webapp/mapsScript.js
```diff
@@ -90,6 +93,93 @@ function showDirections(locations) {
...
+    maxWidth: 200
+  });
+  // Open/close infoWindow on marker click.
+  let infoWindowOpen = false;
+  marker.addListener("click", () => {
+    if (infoWindowOpen) {
+      infoWindow.close();
+      infoWindowOpen = false;
+    }
+    else {
```
adamfweidman, line 173: nit: fix formatting `} else {`

src/main/webapp/mapsScript.js
```diff
@@ -23,6 +23,9 @@ var map;
 var directionsService;
 var directionsRenderer;
 
+// Map marker label constant
+var labels = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ';
```
tristan-jh, line 27: This labeling is implicitly dependent on the fact that maps routing labels it's points as A->Z, instead of, eg: 1->N or some other labeling scheme. That part of the maps api is probably not going to change anytime soon, and I don't think you need to change your implementation, but I think it's still worth noting that this is potentially fragile.

src/main/webapp/mapsScript.js
tristan-jh: I don't think this loop necessarily happens after the directionsRenderer is called. The directionsRenderer is inside of the callback function which could be executed much later.

It seems like it might be working anyways, in which case you may be able to condense the two loops into one.

src/main/webapp/mapsScript.js
```diff
@@ -90,6 +93,108 @@ function showDirections(locations) {
...
+ * isOrigin: boolean indicating if the location is the origin (hotel) or not.
+             Used to determine the icon of the map marker.
+ * index: index of waypoint. Used to determine marker label.
+ */
+function addPoiMarker(location, isOrigin, index) {
+  // Query the location String to get Place details.
+  var service = new google.maps.places.PlacesService(map);
+  service.textSearch({
+    location: map.getCenter(),
+    radius: '500',
```
tristan-jh, line 117: This seems potentially fragile. We certainly don't expect anyone to enter a destination more than 500 (miles?) away, but it seems like they'd get confusing behavior if they did.

More broadly, it seems like doing a textSearch opens the door to various possible edge-cases like: matching the wrong location or matching multiple locations. I'm wondering if it's possible to add this functionality without a textSearch, eg: could you get the place ids from the TripDay and use those to add a map pin?

src/main/webapp/mapsScript.js
eshika: Yes this is true! Ideally I would have liked to just change the origin marker but it seems like the API is structured such that it either automatically adds all or none of the markers and the directionsRenderer is not very transparent, so this is the best workaround I could find.

src/main/webapp/mapsScript.js
eshika: The location/radius parameters in the search only bias the results to prefer locations in the circle, but it will still display locations outside the circle if found. Based on this, I would prefer to leave it in so that the text search is more likely to find the appropriate location we're referring to (e.g. Central Park in NY vs a diff Central Park). 

Ideally we would do one text search in the backend to get the place IDs and then only use the place IDs to refer to the locations from then on to reduce any chance of misrepresenting the location, but that would require more fundamental changes which are not priorities in the interest of time.

src/main/webapp/mapsScript.js
eshika: I played around with combining the loops, but the text search was searching from a center of (0,0) rather than the new center of the map. Added an inline comment explaining why this second loop is needed.

src/main/webapp/mapsScript.js
tristan-jh: That sounds good to me!

PR #42, opened 2020-08-01
Refine the repository setup, remove .html from end of URLs.

zpChris:
### List of changes:
- Move the `Trip.java` and `TripDay.java` files into the data folder with the other java classes (Event, User, and Config).
  - I also changed the imports for those files accordingly.
-  Add the [**UrlRewriteFilter** dependency](https://github.com/paultuckey/urlrewritefilter), `web.xml` file, and `urlrewrite.xml` file in order to remove the `.html` from the end of the URL for the Trips, Maps, and Calendar pages (this will also apply for the Edit page, when that is added).
- Move the Trips files (`trip.html` and `tripScript.js`) into the webapp folder, instead of a separate trips folder.
- Rename the `calendar-script.js` file to `calendarScript.js` to match the formatting of the other JS file names.

### Notes:
- There are many files changed because I had to change the imports of the `.java` files and slightly modified many of the `.html` and `.js` files. In total, the number of changes are relatively small.
- The change to remove the `.html` from the end of the URL wasn't suggested during the User Study, but I wanted to do this when cleaning up the folder structure. In order to do this, I had to use the [**UrlRewriteFilter** dependency](https://github.com/paultuckey/urlrewritefilter). With this, in order to map the servlets to the URLs (for example, `AuthServlet.java` to `/auth`), I had to use the `web.xml` file (see instructions for mapping servlets [here](https://cloud.google.com/appengine/docs/standard/java/config/webxml)).
- In order to get this to work after switching between branches locally, I had to delete the `target` folder to rebuild the site / repo. If it doesn't work, I'd suggest trying this as a first solution.

### Screencast:
No screencast necessary as none of the changes are to the UI.

### Images:
No images necessary as none of the changes are to the UI.

You can see the modifications in the URL; you can access the Trips page directly at `/trips`, the Calendar page at `/calendar?tripKey=TRIP_KEY_HERE`, and the Maps page at `/maps?tripKey=TRIP_KEY_HERE`.


zpChris:

PR #43, opened 2020-08-01
Add Utility classes.

zpChris:
### List of changes:
- Create the `Utility.java` file and include common methods from servlets and Java files.
- Create the `UtilityTest.java` file and include common methods from Java test files.

### Notes:
- The Utility class needed to be separated into `Utility.java` and `UtilityTest.java` for the regular and the testing functions. If all of these methods were put into the `Utility.java` class, an error was thrown -- the mocks, Entity objects, and usage of Datastore needs to be in the test folder.
- Right now, none of the Utility functions are actually used in the rest of the codebase. This CL is expected to be the base CL by which to actually use these Utility methods.
- I'm likely expecting further changes to these files once we start to add these methods to the servlets and other Java classes, but this can be used as a starting point.

### Screencast:
No screencast necessary as none of the changes are to the UI.

### Images:
No images necessary as none of the changes are to the UI.

### DIFF
--- a/src/test/java/com/google/sps/UtilityTest.java
+++ b/src/test/java/com/google/sps/UtilityTest.java
@@ -81,14 +81,13 @@ public class UtilityTest {
   /**
    * Add a Trip Entity to Datastore under the passed-in Key of the User Entity.
    *
-   * @param userEntityKey The Key from the User Entity in Datastore. Must be 
-   * non-null.
-   * @param tripName The human-readable name for the trip. Must be non-null.
-   * @param destinationName The name of the destination the user is heading to.
-   * @param tripKey The key of the trip as created by the Trip Entity.
-   * @param imageSrc The image source / URL to represent the trip.
-   * @param startDate The start date for the trip. Must be non-null. Must be in yyyy-MM-dd date format.
-   * @param endDate The end date for the trip. Must be non-null. Must be in yyyy-MM-dd date format.
+   * @param userEntityKey The Key from the User Entity in Datastore. 
+   * @param tripName The human-readable name for the trip. 
+   * @param destinationName The name of the destination the user is heading to. 
+   * @param tripKey The key of the trip as created by the Trip Entity. 
+   * @param imageSrc The image source / URL to represent the trip. 
+   * @param startDate The start date for the trip. Should be in yyyy-MM-dd date format.
+   * @param endDate The end date for the trip. Should be in yyyy-MM-dd date format.
    */
   public static Entity addTripEntityToDatastoreUnderUser(Key userEntityKey,
     String tripName, String destinationName, String imageSrc, String startDate, 
@@ -170,5 +169,4 @@ public class UtilityTest {
     datastore.put(eventEntity);
     return eventEntity;
   }
-
 }
--- a/src/test/java/com/google/sps/UtilityTest.java
+++ b/src/test/java/com/google/sps/UtilityTest.java
@@ -14,6 +14,7 @@
  
 package com.google.sps.servlets;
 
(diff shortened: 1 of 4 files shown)


src/test/java/com/google/sps/UtilityTest.java
```diff
@@ -0,0 +1,174 @@
...
+    Entity eventEntity = new Entity(Event.QUERY_STRING);
+    eventEntity.setProperty("name", name);
+    eventEntity.setProperty("address", address);
+    eventEntity.setProperty("start-time", 
+        DateTimeFormatter.ISO_LOCAL_DATE_TIME.format(startTime));
+    eventEntity.setProperty("travel-time", travelTime);
+    datastore.put(eventEntity);
+    return eventEntity;
+  }
+
```
eshika, line 173: nit: remove extra line

src/test/java/com/google/sps/UtilityTest.java
```diff
@@ -0,0 +1,174 @@
...
+   * Add a Trip Entity to Datastore under the passed-in Key of the User Entity.
+   *
+   * @param userEntityKey The Key from the User Entity in Datastore. Must be 
+   * non-null.
+   * @param tripName The human-readable name for the trip. Must be non-null.
+   * @param destinationName The name of the destination the user is heading to.
+   * @param tripKey The key of the trip as created by the Trip Entity.
+   * @param imageSrc The image source / URL to represent the trip.
+   * @param startDate The start date for the trip. Must be non-null. Must be in yyyy-MM-dd date format.
+   * @param endDate The end date for the trip. Must be non-null. Must be in yyyy-MM-dd date format.
```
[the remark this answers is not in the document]
zpChris: Removed comments of "Must be non-null" -- thanks!

src/test/java/com/google/sps/UtilityTest.java
zpChris: Done!

zpChris:
> A good start! From what I understand the functions you have in UtilityTest are functions that are repeated in our test files. However, we still need tests for the functions in the Utility class, otherwise this code will go untested when we replace the current functions with Utility functions. I would recommend adding tests for the functions in the Utility class in UtilityTest and creating a separate TestUtils file for the functions that create mocks and set up tests.

I added a `TestUtils.java` file to test the methods in the `Utility.java` class! 

Note: I did not test the `getPlaceDetailsFromPlaceId(...)` method as that does nothing but use an external API, so a test of that would be unnecessary as I'd have to mock that method's objects.

src/test/java/com/google/sps/UtilityTest.java
```diff
@@ -0,0 +1,172 @@
...
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import org.powermock.api.mockito.PowerMockito;
+import org.powermock.core.classloader.annotations.PrepareForTest;
+import org.powermock.modules.junit4.PowerMockRunner;
+
+@RunWith(PowerMockRunner.class)
+@PrepareForTest(UserServiceFactory.class)
+public class UtilityTest {
```
tristan-jh, line 45: I think this class should actually be called TestUtils (or similar) and the file currently called TestUtils should be renamed UtilitiesTest. Elsewhere in this project the convention for test files has been: `<ClassBeingTested>Test.java`. Under that pattern Utilities.java should be tested by UtilitiesTest.java.

TestUtils.java or something similar seems like a fine name for a file which contains util methods used by testing code. The fact that it doesn't end in 'Test.java' is a good indication that it's not a typical unit test file.

zpChris:

PR #44, opened 2020-08-02
Add edit nav button and create base design for the edit page.

zpChris:
### List of changes:
- Add edit navigation buttons to the Calendar, Maps, Trips (and Edit) pages.
- Create the initial design for the edit page with the sidebar.
- Create the `edit.html` and `editScript.js` files; the JavaScript file is used to add the Google Autocomplete widget.

### Notes:
- The "Add POI" button on the Edit page does not work yet; this will be implemented in a future CL.
- Since there isn't enough content in the body of the Edit page, you can't immediately test the "sticky" component of the sidebar because you can't scroll enough; however, adding more content for testing can show that this feature works (see the screencast).
- There's a bug where the Google Places Autocomplete widget for the POI text input becomes misaligned with with text input box. See the [screencast](https://screencast.googleplex.com/cast/NTA4MDQwMjY1NTk2OTI4MHwwNzQzZjgxOC0xOQ) of the bug.
  - I know why this occurs (the widget stays in the same position relative to the site while the sidebar is sticky at the top of the page as you scroll), but I don't know how to fix it. Fixing this issue could be an objective for a future CL, if we have time; that being said, you can easily work around this if you're the user, so it's not a huge bug.
- More content will be added to the Edit page, but in order to keep this CL small I left that out, and will put it on a separate CL. This CL will also be the base branch that other CLs can merge into, since it sets up `edit.html` and `editScript.js`.

### Screencast:
[Edit Page Design and Nav Buttons](https://screencast.googleplex.com/cast/NjI4MTIwMzE2ODU3NTQ4OHwxMDE2MjI4OC1mYw)

### Images:
Image of the Edit page (note: the Point of Interest input field uses the same Google Autocomplete and design as the frontpage input form):
image

Image of the Trips page with the newly-added Edit nav button:
image

Image of the Calendar page with the newly-added Edit nav button:
image

Image of the Maps page with the newly-added Edit nav button (same design as the Calendar page):
image


src/main/webapp/editScript.js
```diff
@@ -0,0 +1,46 @@
...
+  addInputPoiLocationAutofill();
+}
+
+// Add text input POI Google Places autofill.
+function addInputPoiLocationAutofill() {
+  const inputPoi = document.getElementById('inputPoi');
+  let locationAutocomplete = new google.maps.places.Autocomplete(inputPoi);
+
+  // Any time the input changes through user typing, remove the 'is-valid' class
+  // Note: this will not be called if the user clicks on the Google Place autofill.
```
adamfweidman, line 33: nit: change format to 
/** * */
same for comment below

src/main/webapp/edit.html
```diff
@@ -0,0 +1,76 @@
...
+        <a id="button-group-nav-edit">
+          <button type="button" class="btn btn-primary btn-xlarge">Edit</button>
+        </a>
+      </div>
+      <div class="row py-3" id="edit-page-columns">
+        <div class="col-4 order-2">
+          <div class="sticky-top" id="sidebar-container"> 
+            <div class="nav flex-column" id="sidebar-bg">
+              <div class="form-group sidebar-input-form">
+                <label for="inputPoi">Add a Point of Interest</label>
```
adamfweidman, line 59: I'm not sure why you need all of these right now but I assume you will be using this breakup later

src/main/webapp/editScript.js
zpChris: Done!

src/main/webapp/edit.html
zpChris: Yeah, the long list of nested `div`s definitely isn't ideal, but in order to make a two-column structure with a background on the sidebar that rests _slightly below_ the top, this is the only solution I found. If it's possible to make this same code without all the nested `div`s, that'd be ideal, but for now I'm planning to leave it like this.

zpChris:

PR #45, opened 2020-08-03
Add static content to the edit page, create text input with auto-changed size.

zpChris:
(This branch is being merged into **EditPageContent** because that branch has the edit page structure that is needed.)

### List of changes:
- Add content to the sidebar and body (left side) of the Edit page.
  - Add **Save Trip** and **Delete Trip** buttons, as well as the temporarily-added POI button, to the sidebar.
  - Add the trip title, day of travel, and trip cards to the body.
- Create a **Changed** button which will be dynamically added to the page under any edited inputs.
- Create a JavaScript function to have the text inputs match the size of the text. (Note: There is no built-in function to do this, so I had to do it dynamically.)
- Create a card style used for visited POIs with the image on the left, and the text / information on the right.

### Notes:
- Currently, the Edit body content is static / hard-coded. On later CLs, however, this content will be dynamically added.
- The **Changed** button will only appear once the corresponding input element is edited. Right now, however, I added it to two random elements -- the trip title and the second trip card.
- Eventually, the trash can and edit icons will be clickable / actionable.
- When the page is refreshed with a cache-clear (`ctrl + shift + r`), the input text size doesn't match the actual text size. However, on any other situation, the input text size is correct. I don't know what causes this error, but it's fixed as soon as the user modifies the text.

### Screencast:
[Add Edit Page Content](https://screencast.googleplex.com/cast/NTE4NDYwODQ5NDY4MjExMnw3YWExOTMwMS0zZg)

### Images:
Image of the Edit page:
image

### DIFF
--- a/src/main/webapp/edit.html
+++ b/src/main/webapp/edit.html
@@ -123,7 +123,7 @@
                   <input class="autoscale-text edit-card-title-size" type="text" value="Woodland Park Zoo">
                   <i class="fa fa-edit edit-card-icon"></i>
                   <p class="edit-card-text">5500 Phinney Avenue North, Seattle</p>
-                  <p class="edit-card-text">10AM - 11AM</p>
+                  <p class="edit-card-text">11:10AM - 12:10PM</p>
                 </div>
                 <i class="fa fa-trash edit-card-delete"></i>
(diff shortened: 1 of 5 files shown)


zpChris:

PR #48, opened 2020-08-04
Correct redirects to catch various error cases.

zpChris:
### List of changes:
- On the Calendar and Maps pages, the site will redirect to the `/trips/` page if the `tripKey` is not present or is invalid.

### Notes:
- The JavaScript file catches an error when the `fetch(...)` call finds a redirect, but the site still works as expected because the URL is redirected.
- The "try / catch" blocks in the Java servlets are to catch when `KeyFactory.stringToKey(...)` throws an `IllegalArgumentException` when presented with an invalid String representation of a key.

### Screencast:
[Correct Redirects on Calendar and Maps](https://screencast.googleplex.com/cast/NjI0MTQ4NjkzNDkwMDczNnxhM2NjODczNC1mOQ)

### Images:
There are no images in this CL because there are no changes to the UI, just redirects.


zpChris:

PR #51, opened 2020-08-06
Create EditServlet and add tests.

zpChris:
### List of changes:
- Get the Trip, TripDay, and Event objects from Datastore in a properly-formatted manner for the Edit page.
  - The properly-formatted manner is a JSON object that stores the overall Trip data, and a map between each date of the trip and the events on that day (ordered from start to end date, and events ordered from first to last).
  - Add error handling if the necessary objects are not in Datastore.
- Add unit tests for the `EditServlet.java` file (`EditServletTest.java`).

### Notes:
- This CL does not have many integration tests, specifically for `doGet(...)` and `writeEditTripContentToFile(...)` in the `EditServlet.java` file, due to time constraints from the end of the internship -- this was approved by my hosts.
- The `LinkedHashMap` preserves its order when converted to JSON through Gson; see [this discussion](https://github.com/googleinterns/step135-2020/pull/51#discussion_r466728133) for more information.

### Screencast:
There are no screencasts in this CL because there are no changes to the UI, just retrieving objects from Datastore.

### Images:
There are no images in this CL because there are no changes to the UI, just retrieving objects from Datastore.


src/main/java/com/google/sps/servlets/EditServlet.java
```diff
@@ -0,0 +1,246 @@
...
+    if (tripDayList.isEmpty()) {
+      response.sendRedirect("/trips/");
+      return;
+    }
+
+    /**
+     * Get the events, and put them in a map relating the date to the Event list.
+     * The LinkedHashMap is used to maintain the ordering of the TripDays.
+     */
+    Map<String, List<Event>> dateEventMap = new LinkedHashMap<>();
```
tristan-jh, line 118: No change requested. I'm curious if converting the EditTrip object into json preserves the order of the map keys, and if that order can be used in the frontend code. Doing a quick google it looks like iteration order of json object keys in javascript is _complicated_. I'm fine with this as is, but if converting to json then parsing in javascript shuffles the order of keys in a map, then this servlet wouldn't need to bother sorting TripDays since they'll just be shuffled later anyways.

If we had time this would be something to manually test and consider changing, but for now I think the attempt to provide TripDays in chronological order makes a lot of sense.

src/test/java/com/google/sps/EditServletTest.java
```diff
@@ -0,0 +1,520 @@
...
+  private EditServlet editServlet;
+
+  // Add constants for testing User class.
+  public static final String EMAIL = "testemail@gmail.com";
+
+  // Constants to represent different TripDay attributes.
+  private static final String INPUT_DESTINATION = 
+    "Space Needle, Broad Street, Seattle, WA, USA";
+  private static final String INPUT_DESTINATION_2 = 
+    "Woodland Park Zoo, Phinney Avenue North, Seattle, WA, USA";
```
tristan-jh, line 63: Both of these are duplicated below with SPACE_NEEDLE_ADDRESS and WOODLAND_ADDRESS respectively. I think it's fine to use the below constants wherever you're currently using these.

src/test/java/com/google/sps/EditServletTest.java
```diff
@@ -0,0 +1,520 @@
...
+      LocalDateTime.of(LocalDate.parse("2020-08-22"), LocalTime.of(10, 00));
+      private static final String WOODLAND_PLACE_ID = "5678";
+  private static final int HALF_HOUR = 30;
+
+  private final LocalServiceTestHelper helper =
+      new LocalServiceTestHelper(new LocalDatastoreServiceTestConfig());
+
+  @Before
+  public void initEditServlet() {
+    editServlet = new EditServlet();
```
tristan-jh, line 95: Do these tests work if you move this line to setUp()? It's stylistically fine to do all the initialization in a single `@Before` method.

src/test/java/com/google/sps/EditServletTest.java
```diff
@@ -0,0 +1,520 @@
...
+   * @param tripName The name of the trip.
+   * @param destinationName The name of the trip destination.
+   * @param imageSrc The String URL of the image that represents the trip.
+   * @param tripStartDate The String start date of the trip. Must be in 
+      "yyyy-mm-dd" format.
+   * @param tripEndDate The String start end of the trip. Must be in 
+      "yyyy-mm-dd" format.
+   * @return The Trip Entity that was placed in Datastore.
+   */
+  public Entity createTripEntity(DatastoreService datastore, Entity userEntity,
```
tristan-jh, line 147: I'd actually be fine with it taking just the datastore and userEntity as parameters and hard-coding all other properties of the trip. That would be poor form outside the context of a test, but here it might be fine because 1) this method is always called with the same arguments, and 2) there are no assertions about the properties of the trip. In essence this method is used to _put some trip in the datastore_ but we never care about the exact properties of the trip which is put there. But like Bootsie said this morning, this is more a matter of preference than strict style guidance. If you feel that having the parameters makes the tests more readable, then there's no need to change. I think there's good arguments either way.

src/main/java/com/google/sps/servlets/EditServlet.java
```diff
@@ -0,0 +1,246 @@
...
+import javax.servlet.http.HttpServletResponse;
+
+@WebServlet("/get-edit-content")
+public class EditServlet extends HttpServlet {
+
+  private final String TRIP_KEY_PARAM = "tripKey";
+  private final String KEY_FILTER_NAME = "__key__";
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response)
```
tristan-jh, line 55: Not requesting a change but I wanted to connect this back to the design discussion we had on Tuesday. Like we talked about then one option for the overall design of this project would be to consolidate the different get get/post endpoints to reduce overlap. I think this doGet function returns a superset of the information needed by the calendar page. So an example of how to do that consolidation would be to update the calendar js page to  use this endpoint and filter through the returned EditTrip object to get the date and event information it needs.

Like we talked about then, this is just one strategy for the high-level design of this project. And even if that was the design we decided to follow, such changes are definitely out of scope for this CL.

src/main/java/com/google/sps/data/Event.java
```diff
@@ -175,4 +174,9 @@ public LocalDateTime getEndTime() {
   public long getTravelTime() {
     return this.travelTime;
   }
+
```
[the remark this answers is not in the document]
zpChris: Done!

src/main/java/com/google/sps/servlets/EditServlet.java
```diff
@@ -0,0 +1,246 @@
...
+    return tripEntity;
+  }
+
+  /**
+   * Get the TripDay Entity objects from datastore that are under the passed-in
+   * Trip Entity Key.
+   * 
+   * @param tripKey The Key of the trip used as an ancestor of the TripDay objects.
+   * @param datastore The datastore object to use when searching for the TripDay objects.
+   * @return The List of TripDay Entity objects under the passed-in Trip Key.
```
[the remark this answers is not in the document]
zpChris: Done!

src/main/java/com/google/sps/servlets/EditServlet.java
zpChris: Thanks for asking this -- this **does** preserve the order when converting to JSON! This is why I used `LinkedHashMap` instead of a simple `HashMap`; as discussed on this [Stack Overflow answer](https://stackoverflow.com/questions/29491281/building-ordered-json-string-from-linkedhashmap), GSON preserves the current order of the keys when converting from a `LinkedHashMap`.

This isn't tested on this CL (see Notes at the top of the CL, there were time constraints to testing), but on the branch used for our final presentation ("`FinalPresentation`"), I verified that this Map maintains its sorting order when converted to JSON.

src/test/java/com/google/sps/EditServletTest.java
zpChris: Done!

src/test/java/com/google/sps/EditServletTest.java
zpChris: Ok, that makes sense -- fixed!

src/test/java/com/google/sps/EditServletTest.java
zpChris: Yeah, that makes sense. I'm going to leave this currently as-is because future tests could require building of two different Trip objects (such as the current tests in this CL with two TripDay or two Event objects), and because I'd prefer to maintain uniform design decisions throughout the helper methods (the `createTripDayEntity(...)` method requires taking parameters). Thanks for this note 👍

src/main/java/com/google/sps/servlets/EditServlet.java
zpChris: That all makes sense -- thanks for including these details 🙂

PR #52, opened 2020-08-06
Algorithm to Solve TSP (Traveling Salesman Problem) with Time Windows 

adamfweidman:
[Design Doc is linked to clarify algorithm](https://docs.google.com/document/d/1SxCdIYOzTooS1ak_l5J1SpRlsNWwijmuzpjq_58i4zo/edit#)

### Files Changed:

**TspSolver.java**: 
- Responsible for solving and setting a variable `finalAnswer`that contains the total travel time (and wait time if applicable) and a list of ints (representing placeIds) in the determined correct order.
- Purpose: The current method of finding the most efficient route does not take into account business hours of a location. This algorithm takes into account that a user is not able to visit a location unless it is open.

Algorithm basic outline:
- Set a `timeMatrix` holds the travel time between all locations, `intToPlaceId` map between ints and placeIds, and finally opening hours indexed at corresponding int representation of placeIds
- Recursive part of algorithm that keeps track of the lowest cost and path associated with that cost. Makes sure that any place visited is either currently open or will be open (wait time incorporated in the cost)
- Note: format of opening hours returned by the placesAPI is often irregular, many checks are implemented to see if they exist for desired day. However, when opening hours were not found they were set to be open "All Day".

**TspSolverTest.java**
- Described in the file itself but to reiterate: Mocking the google API objects was not done because of the time constraints so the `createTimeMatrix()` and `populateIntMapAndOpenHours()` functions were not tested. 
- Currently tested is the recursive part of the algorithm `solverHelper()`. In order to just test this method there are many methods in the test class exclusively responsible for manually creating opening hours, int to placeId HashMap, and the timeMatrix.
- `testBasicRecursiveAlgorithmSolver()` tests algorithmic functionality where travel time is the only factor
- `testAlgorithmHoursDetermine()`: tests algorithmic functionality taking into account travel time and opening hours
- `testAlgorithmReturnsEmpty()` tests that the algorithm returns an empty list when the trip is not possible (view all locations when open)

**TripServlet.java**
- Change scope of a variable and a method used in  `TspSolver.java`


src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+  }
+
+  /**
+   * Check if a location is or will be open. Possibilities:
+   * 1. Arrival before open time 
+   * 2. Arrival during open hours
+   * 3. Arrival after closed 
+   * 4. Open Hours are not available
+   * 
+   * @return List<Object> of length 2
```
tristan-jh, line 196: change requested: mixing types in List is usually discouraged.

A more typical way to do this would be to just return an int but have a "sentinel value" to indicate case 3. Eg: either return the time added (some number > 0) or returns -1 if arriving after close. This makes it easier for the code that calls this method because it doesn't have to deal with casting elements from the returned List<Object>.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+   * @param index index to put result in OpenHours Map
+   */
+  private void checkDayIsPresentInHours(OpeningHours openingHours, int index) {
+      int adjustedWeekIndex;
+      // periods is numbered Sunday - Saturday, WeekdayText is Monday - Sunday
+      if (intOfWeek == 0) {
+        adjustedWeekIndex = 6;
+      } else {
+        adjustedWeekIndex = intOfWeek - 1;
+      }
```
tristan-jh, line 337: totally optional, but check out the modulus operator (`%`) for an alternate way to do this. Eg: `(intOfWeek + 6) % 7`

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+      String desiredDayString = WEEKDAYS[adjustedWeekIndex];
+      for (int i = 0; i < openingHours.periods.length; i++) {
+        // get the day that the ith index in opening hours corresponds to
+        String day = openingHours.weekdayText[i].split(":")[0];
+
+        // check neither open nor closing time are null
+        boolean open = openingHours.periods[i].open != null;
+        boolean close = openingHours.periods[i].close != null;
+
+        if (day.equals(desiredDayString) && open && close) {
```
tristan-jh, line 348: nit: it seems like day.equals(desiredDayString) is always going to be true? If so you might not need the WEEKDAYS constant.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+  private static final int START_COST = 0;
+  private static final int START_ANS = Integer.MAX_VALUE;
+  private static final LocalTime START_TIME = LocalTime.of(10, 0);
+  private static final int HOUR = 60;
+
+  private GeoApiContext context;
+
+  // variables to create before alg
+  int[][] timeMatrix;
+  HashMap<Integer, String> intToPlaceId;
```
tristan-jh, line 62: nit: Map<Integer,String>

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+        
+        // edge case where close time is next day at 0:00
+        if (close.compareTo(LocalTime.parse("00:00")) == 0) {
+          close = LocalTime.of(23, 59);
+        } else {
+          close.minusMinutes(HOUR);
+        }  
+
+        // set time after travelTime
+        LocalTime afterDrive = currentTime.plusMinutes(timeMatrix[currPos][i]);
```
tristan-jh, line 225: I was a little surpized to find this logic here.

I think it's correct, but I'd argue that might be clearer to rename this method somewhat. Eg: rename this method to be something like "isReachableBeforeClosing" or "travelPlusWaitTime", then rename the arguments to be departureTime, departurePoi, targetPoi or similar. That way it's clear that this method is accounting for the travel time.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+   * Recursive helper function that is used to solve computational part of
+   * algorithm. Called in global `solver()` method
+   *
+   * @param currPos represents the placeId of where the alg currently is
+   * @param numNodes total number of locations to visit (includes center)
+   * @param currentTime time if path at this point were taken 
+   * @param count number of pois visited during this path
+   * @param cost Tuple of current path (time cost, List of pois)
+   * @param ans the current best solution found
+   * @param array of which pois have been visited by this path
```
tristan-jh, line 135: optional: these descriptions are pretty clear. But I think it would still be easier to understand this method if it had fewer arguments. I think you can actually skip some of these arguments without too much trouble.

Eg: currPos will always be equal to the final element in `cost.getCurrPath()`. Similarly count is equal to `cost.getCurrPath().size()` and I think numNodes will always be equal to `visited.length`.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+      if (!visited[i] && timeMatrix[currPos][i] > 0 && isOpen) {
+        // mark ith node as visited
+        visited[i] = true;
+
+        // update cost
+        LocalTime timePostUpdate = currentTime.plusMinutes((long) timeToAdd);
+        List<Integer> costPath = cost.getCurrPath();
+        List<Integer> updatePath = new ArrayList<>();
+        updatePath.addAll(costPath);
+        updatePath.add(i);
```
tristan-jh, line 174: This is fine. Totally optional, but another way you can do this would be to add directly to cost.getCurrPath() here then remove the element you added from cost.getCurrPath below where you reset visited. This might make the code a little harder to reason about, but it reduces the storage needs of this algorithm since you'd only have 1 list instead of 1 list per recursive call.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+    DirectionsResult dirResult = TripServlet.getDirectionsResult(directionsRequest);
+
+    // get the time in minutes of the first the trip
+    DirectionsLeg leg  = dirResult.routes[TripServlet.ROUTE_INDEX].legs[0];
+    int travelTime = (int) leg.duration.inSeconds / TripServlet.SECONDS_IN_MIN;
+    return travelTime;
+  }
+
+  /**
+   * Getter functions
```
tristan-jh, line 381: nit: might be worth mentioning that all of these except for getFinalAnswer are for tests only.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+    for (int i = 0; i < allPois.size(); i++) {
+      timeMatrix[i][i] = 0;
+    }
+
+    // fill (i, j) and (j, i) indices with travel time from i to j
+    for (int i = 0; i < allPois.size(); i++) {
+      for (int j = i; j < allPois.size(); j++) {
+        int travelTime = getTravelTimeMins(allPois.get(i), allPois.get(j));
+        timeMatrix[i][j] = travelTime;
+        timeMatrix[j][i] = travelTime;
```
tristan-jh, line 279: Nice! I like you use the symetric nature of travel times to only query i->j then fill in both the [i][j] and [j][i] locations in the matrix.

I'm not sure but the DistanceMatrixApi might let you get all the values you need in a single request which would improve the performance here even more: https://github.com/googlemaps/google-maps-services-java/blob/master/src/main/java/com/google/maps/DistanceMatrixApi.java

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    /**
+     * Backtracking Step: loop through the time Matrix from currPos, increase
+     * count by 1 and update (int) and (List) value of cost
+     */
+    for (int i = 0; i < numNodes; i++) {
+      // set open boolean and time to add to currenTime
+      boolean isOpen = (boolean) checkIsOpenSetTime(currentTime, i,
+          currPos).get(0);
```
tristan-jh, line 160: meant to include this in the review but this check is great because it lets you quit early as soon as your route arrives somewhere after closing. so, eg: if the second destination would be closed when you arrive this algorithm doesn't waste any time considering what would happen after that in the route. Nice!

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: Right! This is much better: it felt weird/wrong to have a list of different objects.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: It won't always return true because I'm checking that the location is open for the specific day I'm looking for. So if I'm looking for open hours of "Tuesday", `desiredDayString` will be "Tuesday" but if `i=0` then `day` will be "Monday" and the if statement will be entered and put in wrong hours (Tuesday may not have any open hours).

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: I should have thought about using mod that's much cleaner!

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: I had issues in the past about sending the same object through the recursive call multiple times. I think I'm going to leave as is.

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: The issue with this is that the first iteration has size=0 and then currPos is trying to get index 0 from an empty list

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: I understand that I could use this however this has calculates all times between locations (from A to B, B to A, and from A to A). For all pairs of locations. Requesting from the server may take longer but it will also take restructuring of the `createTimeMatrix()` method so i will keep it as is.

(current function makes `n^2 / 2 - n` requests. Given that n will never  be too large since its the number of locations per day I believe it's reasonable to keep as is)

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: Fixed by adding 0 (index of central location) into the list at the start of the recursion

src/main/java/com/google/sps/data/algorithm/TspSolver.java
adamfweidman: Agreed! The names of method and variables have been changed

src/main/java/com/google/sps/data/algorithm/TspSolver.java
```diff
@@ -156,10 +162,8 @@ public Tuple solverHelper(int currPos, int numNodes, LocalTime currentTime,
      */
     for (int i = 0; i < numNodes; i++) {
       // set open boolean and time to add to currenTime
-      boolean isOpen = (boolean) checkIsOpenSetTime(currentTime, i,
-          currPos).get(0);
-      int timeToAdd = (int) checkIsOpenSetTime(currentTime, i,
-          currPos).get(1);
+      boolean isOpen = travelPlusWaitTime(currentTime, i, currPos) != -1;
+      int timeToAdd = travelPlusWaitTime(currentTime, i, currPos);
```
tristan-jh, line 166: nit: just call this method one time then check if timeToAdd is -1 afterwards. Saves the trouble of computing the value twice.

adamfweidman: Done!

